PR #5, opened 2016-10-12
Added support for API calls

devrobo:


test/system/test_cvp_client.py
```diff
@@ -31,6 +31,13 @@
 #
 
 ''' System test for the CvpClient class
+
+    Requirements for CVP Node:
+    1) Just need one node for test with the following account:
```
jerearista: Can these be (optionally) supplied another way instead of being hard coded?  I'm thinking of something like `eapi.conf`.  A `.cvprc`, perhaps?  cvprac as a whole might benefit from that.  I could add that as an enhancement request instead of holding this up.

cvprac/cvp_api.py
```diff
@@ -0,0 +1,500 @@
...
+        ''' Initialize the class.
+
+            Args:
+                clnt (obj): A CvpClient object
+        '''
+        self.clnt = clnt
+        self.log = clnt.log
+
+    def get_cvp_info(self):
+        ''' Initialize the class variables.  This routine is called from the
```
devrobo: No it is not, I will update.

test/system/test_cvp_api.py
```diff
@@ -0,0 +1,395 @@
...
+        result = self.api.search_topology(self.device['fqdn'])
+        self.assertIsNotNone(result)
+        dev_container = result['netElementContainerList']
+        self.assertGreaterEqual(len(dev_container), 1)
+        info = dev_container[0]
+        result = self.api.search_topology(info['containerName'])
+        self.assertIsNotNone(result)
+        self.container = result['containerList'][0]
+
+        # Get the configlets assigned to the device.  There must be at least 1.
```
devrobo: You really have to have at least one configlet that sets up AAA and management IP address. If I create one then I have to set that up otherwise applying the configlet would fail because it would be wiping out the user account and management IP on the switch. The tests give an accurate error for missing configlet or node but not for incorrect user credentials. I will add a check at setup for that.

test/system/test_cvp_client.py
devrobo: cvprac, like pyeapi and rbeapi, are a component of something bigger. I did not go the .conf route instead letting the application solve this problem. For the ServiceNow CVP integration there is a .conf file for the integration that contains ServiceNow and CVP login information. For the Monitor and Management app the creds come from logging into the app.

cvprac/cvp_api.py
```diff
@@ -0,0 +1,500 @@
...
+            Returns:
+                tasks (dict): The 'total' key contains the number of tasks,
+                    the 'data' key contains a list of the tasks.
+        '''
+        self.log.debug('get_tasks:')
+        data = self.clnt.get('/task/getTasks.do?queryparam=&startIndex=%d&'
+                             'endIndex=%d' % (start, end))
+        return data
+
+    def get_logs_by_id(self, task_id, start=0, end=0):
```
grybak-arista: Default value for start not documented.

cvprac/cvp_api.py
```diff
@@ -0,0 +1,500 @@
...
+
+            Returns:
+                history (dict): The configlet dict with the changes from
+                    most recent to oldest.
+        '''
+        self.log.debug('get_configlets_history: key: %s' % key)
+        return self.clnt.get('/configlet/getConfigletHistory.do?configletId='
+                             '%s&queryparam=&startIndex=0&endIndex=0' % key)
+
+    def get_inventory(self):
```
grybak-arista: Would there be any reason to implement the start and end values as optional rather than fixed at 0?

cvprac/cvp_api.py
```diff
@@ -0,0 +1,500 @@
...
+            Args:
+                container_name (str): Container name
+                parent_name (str): Parent container name
+                parent_key (str): Parent container key
+        '''
+        self.log.debug('add_container: container: %s parent: %s parent_key: %s'
+                       % (container_name, parent_name, parent_key))
+        self._container_op(container_name, '', parent_name, parent_key, 'add')
+
+    def delete_container(self, container_name, container_key, parent_name,
```
grybak-arista: Missing container_key documentation

test/system/test_cvp_client.py
devrobo: The error is obvious is the CVP username or password are not correct. If the device does not have username and password the same as the CVP user used then the error is not obvious as it is found in the task log. Documented this case.

test/system/test_cvp_api.py
devrobo: The error is obvious is the CVP username or password are not correct. If the device does not have username and password the same as the CVP user used then the error is not obvious as it is found in the task log. Documented this case.

cvprac/cvp_api.py
devrobo: Great catch! Will fix.

cvprac/cvp_api.py
devrobo: Will fix

test/system/test_cvp_client.py
jerearista: Sounds fine.  Just making sure we had the conversation.

cvprac/cvp_api.py
```diff
@@ -0,0 +1,500 @@
...
+                 'fromId': '',
+                 'nodeName': '',
+                 'fromName': '',
+                 'toName': dev['fqdn'],
+                 'childTasks': [],
+                 'parentTask': ''}]
+        self.log.debug('remove_configlets_from_device: saveTopology data:\n%s'
+                       % data)
+        self.clnt.post('/provisioning/saveTopology.do', data)
+
```
devrobo: I am not seeing a problem here on 2016.1.2.1. In the applicable cvprac test case it adds and removes a trivial configlet that only changes one line. This works fine without setting the "node*" parameters. I checked if there is an issue if the configlet changes the management IP address and ran that test and it worked fine without setting the "node*" parameters. I verified that CVP automatically picked up the new management address for the node.

I wonder if the "node*" parameters need to be set when ZTPing a new node. If that is the case then I can add them as optional parameters to the method call.

Issue #7, opened 2016-11-22
Create routine to allow setting the get and post default timeout value


devrobo:
Currently a user of the API modules has no way to set the timeout on a per request basis which means that the API calls will always use the default timeout value. This call allows the user to change the timeout on the fly.

PR #12, opened 2016-12-14
Added request timeout instance variable to API.

mharista:
Allows user to specify client get/post timeouts used by API calls. Issue #7

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -61,7 +61,7 @@ class CvpApi(object):
     '''
     # pylint: disable=too-many-public-methods
 
-    def __init__(self, clnt):
+    def __init__(self, clnt, request_timeout=30):
         ''' Initialize the class.
 
             Args:
@@ -69,6 +69,7 @@ class CvpApi(object):
         '''
         self.clnt = clnt
         self.log = clnt.log
+        self.request_timeout = request_timeout
 
     def get_cvp_info(self):
         ''' Returns information about CVP.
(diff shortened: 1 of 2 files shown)


mharista:
I had a discussion with Cheyne about this change and he mentioned doing this in a way where the new timeout is a part of the CvpClient instead of CvpApi. This would require less code changes and keep the CvpApi class agnostic to any timeout/connection parameters. A quick usage example is below:

clnt = CvpClient()
clnt.connect(....)
clnt.response_timeout = 40
clnt.api.get_cvp_info()
clnt.response_timeout = 30

I decided to add the functionality to CvpApi because it seems more intuitive to me at have a timeout for api calls accessed via the api. It also makes use of the get and post client functions already accepting a timeout parameter. Same code with one difference is below:

clnt = CvpClient()
clnt.connect(....)
clnt.api.response_timeout = 40
clnt.api.get_cvp_info()
clnt.api.response_timeout = 30

I don't have a preference either way, but I'm curious about others preferences.

devrobo:
You are right, in looking at the example code it is more intuitive to handle the timeout in the api code.

cvprac/cvp_api.py
```diff
@@ -61,22 +61,23 @@ class CvpApi(object):
...
+        self.request_timeout = request_timeout
 
     def get_cvp_info(self):
         ''' Returns information about CVP.
 
             Returns:
                 cvp_info (dict): CVP Information
         '''
-        return self.clnt.get('/cvpInfo/getCvpInfo.do')
+        return self.clnt.get('/cvpInfo/getCvpInfo.do', self.request_timeout)
```
devrobo: The timeout parameter is a named parameter so I think the syntax should be:
timeout=self.request_timeout

devrobo: The timeout parameter is a named parameter so I think the syntax should be:
timeout=self.request_timeout . I think this should be done for all calls below.

PR #18, opened 2017-01-31
Make REST call to addTempAction before saveTopology.

### LINKED ISSUE #17: addTempAction call required before saveTopology in newer releases.
When doing add and delete containers in 2016.1.2.1 the new version requires a 'provisioning/addTempAction.do' call before the '/provisioning/saveTopology.do' call, but the current methods only do the saveTopology.do.

mharista:
Fixes #17


mharista:
Verified all system tests pass old and newer versions of CVP.

PR #23, opened 2017-02-22
Added base deploy_device function.

### LINKED ISSUE #22: add deploy_device functionality
it would simplify the cron script needed for the [Ansible|Puppet|Chef] <--> CVP integration

mharista:
Fixes #22

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -410,6 +410,8 @@ class CvpApi(object):
                 app_name (str): The application name to use in info field.
                 dev (dict): The switch device dict
                 new_configlets (list): List of configlet name and key pairs
+                create_task (bool): Determines whether or not to execute a save
+                    and create the tasks (if any)
 
             Returns:
                 response (dict): A dict that contains a status and a list of
@@ -475,6 +477,8 @@ class CvpApi(object):
                 app_name (str): The application name to use in info field.
                 dev (dict): The switch device dict
                 del_configlets (list): List of configlet name and key pairs
+                create_task (bool): Determines whether or not to execute a save
+                    and create the tasks (if any)
 
             Returns:
                 response (dict): A dict that contains a status and a list of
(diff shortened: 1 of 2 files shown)


cvprac/cvp_api.py
```diff
@@ -752,9 +756,10 @@ def apply_image_to_container(self, image, container):
                     Ex: {u'data': {u'status': u'success', u'taskIds': [u'32']}}
         '''
         return self.apply_image_to_element(image, container, container['name'],
-                                           'container')
+                                           'container', create_task)
 
-    def apply_image_to_element(self, image, element, name, id_type):
+    def apply_image_to_element(self, image, element, name, id_type,
+                               create_task):
```
dathelen: this is probably being a bit picky but should you set the default value of create_task = True here just in case this method is called directly so it will match the other edits?

cvprac/cvp_api.py
```diff
@@ -827,3 +833,60 @@ def remove_image_from_container(self, image, container):
...
+                                  'netElementId=%s' % device['key'])
+        new_configlets = prop_conf['proposedConfiglets']
+        if configlets:
+            new_configlets.extend(configlets)
+        self.apply_configlets_to_device('deploy_device', device,
+                                        new_configlets, create_task=False)
+
+        # Apply image to the device
+        if image:
+            self.apply_image_to_device(image, device, create_task=False)
```
dathelen: Just curious.  So the image argument passed in here needs to an image bundle object?  What happens if I just send in an image name here?  Does it just fall through and cvprac will raise an exception?

mharista: Yes as of now a key value error will be raised when apply_image_to_device attempts to reference a key in a non dict object. I did this to be consistent with how the image is passed as its information dict to the other image functions. I'm not opposed to changing to handle the name instead of the dictionary.

cvprac/cvp_api.py
```diff
@@ -246,20 +246,19 @@ def get_inventory(self, start=0, end=0):
...
-        self.log.debug('get_device_container_map: called')
+        self.log.debug('get_devices_in_container: called')
         data = self.clnt.get('/inventory/getInventory.do?'
-                             'queryparam=&startIndex=%d&endIndex=%d' %
-                             (start, end), timeout=self.request_timeout)
-        return data['containerList']
+                             'queryparam=%s&startIndex=0&endIndex=0' % name,
+                             timeout=self.request_timeout)
+        if data['total'] > 0:
+            return data['containerList']
```
dathelen: Per our discussion I think it would be more helpful to return the netElementList here.

PR #26, opened 2017-03-23
Add ability for user to configure log level.

### LINKED ISSUE #21: Add user-selectable option to set the logging level
Suggest making this an option when instantiating the class and also allowing it to be modified afterward so a developer can increase or decrease the logging level just during a specific section of their code.

mharista:
Fixes #21

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -376,6 +376,9 @@ class CvpApi(object):
                 config (str): Switch config statements
                 key (str): Configlet key
                 name (str): Configlet name
+
+            Returns:
+                data (dict): Contains success or failure message
         '''
         self.log.debug('update_configlet: config: %s key: %s name: %s' %
                        (config, key, name))
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -382,8 +382,8 @@ class CvpApi(object):
 
         # Update the configlet
         body = {'config': config, 'key': key, 'name': name}
-        self.clnt.post('/configlet/updateConfiglet.do', data=body,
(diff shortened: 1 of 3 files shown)


cvprac/cvp_api.py
```diff
@@ -382,8 +382,8 @@ def update_configlet(self, config, key, name):
 
         # Update the configlet
         body = {'config': config, 'key': key, 'name': name}
-        self.clnt.post('/configlet/updateConfiglet.do', data=body,
-                       timeout=self.request_timeout)
+        return self.clnt.post('/configlet/updateConfiglet.do', data=body,
```
cheynearista: Update func header to include what's being returned.

cvprac/cvp_client.py
```diff
@@ -111,7 +111,8 @@ class CvpClient(object):
     # CVP node.
     NUM_RETRY_REQUESTS = 3
 
-    def __init__(self, logger='cvprac', syslog=False, filename=None):
+    def __init__(self, logger='cvprac', syslog=False, filename=None,
+                 log_level='INFO'):
```
cheynearista: Do we need to pass log_level in at init if we provide a set method? :)

mharista: I'm fine with leaving it in both places or removing it from init. I put it in both places because that is how the issue was worded.

cvprac/cvp_api.py
mharista: Done.

cvprac/cvp_client.py
cheynearista: Fair enough

PR #27, opened 2017-03-28
Add API function for validate_config

jnewbigin:


jnewbigin:
I will try and get our lab environment set for this so I can write/run some tests

mharista:
Hi @jnewbigin,

This is a useful module to had added to the API and I have some time that I can use to add in the test cases. I also have a test environment. If you already have the test cases let me know, otherwise I'll merge this in as is.

jnewbigin:
Hi @mharista,
Thanks for merging this! 
We are commissioning some new kit and looking to use some more features of CVP, which probably means some more new functions.
It would be great for us to get out test environment set up. Are there some docs you can point me to for that?
Thanks

PR #33, opened 2017-05-15
Update client to default to https with http fallback.

### LINKED ISSUE #28: Change default transport to HTTPS
Now that CVP prefers HTTPS for transport, that should be the default for the library.  A clear error message should be raised when we fail to connect over HTTPS to help anyone using older versions of CVP who get surprised by not seeing the release notes.

Optionally, default to HTTPS with automatic fallback to HTTP.  If implemented, this should ONLY occur on the initial connect, save the working transport and NOT try to fallback on subsequent requests.  If we successfully connect using the fallback, a warning should be generated.  Fallback must have a disable option.  One could decide to disable fallback if HTTPS is intentionally set on the connect instead of allowing the defaults to fall through for that.

mharista:
Fixed system tests and added some client unit tests.

Fixes #28

### DIFF
--- a/cvprac/cvp_client.py
+++ b/cvprac/cvp_client.py
@@ -126,6 +126,7 @@ class CvpClient(object):
                 log_level (str): Log level to use for logger. Default is INFO.
         '''
         self.authdata = None
+        self.cert = False
         self.connect_timeout = None
         self.cookies = None
         self.error_msg = ''
@@ -170,7 +171,7 @@ class CvpClient(object):
         self.log.setLevel(getattr(logging, log_level))
 
     def connect(self, nodes, username, password, connect_timeout=10,
-                protocol='http', port=None):
+                port=None, cert=False):
         ''' Login to CVP and get a session ID and cookie.  Currently
             certificates are not verified if the https protocol is specified. A
             warning may be printed out from the requests module for this case.
@@ -181,12 +182,13 @@ class CvpClient(object):
                 password (str): The CVP password
                 connect_timeout (int): The number of seconds to wait for a
                     connection.
-                protocol (str): The type of protocol to use for the connection.
-                    The default value is 'http'.
                 port (int): The TCP port of the endpoint for the connection.
                     If this keyword is not specified, the default value is
                     automatically determined by the transport type.
                     (http=80, https=443)
+                cert (str): Path to a cert file used for a https connection.
+                    If a cert is provided then the connection will not attempt
+                    to fallback to http.
 
             Raises:
                 CvpLoginError: A CvpLoginError is raised if a connection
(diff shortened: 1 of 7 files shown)


cvprac/cvp_client.py
```diff
@@ -170,7 +171,7 @@ def set_log_level(self, log_level='INFO'):
         self.log.setLevel(getattr(logging, log_level))
 
     def connect(self, nodes, username, password, connect_timeout=10,
-                protocol='http', port=None):
+                port=None, cert=False):
```
cheynearista: I have a love hate relationship with python. Dynamic typing is nice, but in this case it's confusing. When I see cert=False, I think cert is a bool...but cert can also be a path to cert file. So maybe update the comment to reflect that this could be either a boolean or a string.

cvprac/cvp_client.py
```diff
@@ -234,9 +227,19 @@ def _create_session(self, all_nodes=False):
...
+            self.url_prefix = ('https://%s:%d/web' % (host, self.port or 443))
             error = self._reset_session()
+            if error and not self.cert:
+                self.log.warning('Failed to connect over https. Potentially'
+                                 ' due to an old version of CVP. Attempting'
+                                 ' fallback to http. Error: %s' % error)
+                # Attempt http fallback if no cert file is provided. The
+                # intention here is that a user providing a cert file
+                # forces https.
+                self.url_prefix = self.url_prefix.replace('https', 'http', 1)
```
cheynearista: This could be simplified a little. No need to replace:
self.url_prefix = ('http://%s/web' % (host))

cvprac/cvp_client.py
mharista: I could change the default for cert to None. None or path to file would make a bit more sense than bool or path to file. I would just have to test this to verify I can pass None as the verify parameter to the requests get method.

cvprac/cvp_client.py
mharista: I am still allowing the user to provide a port number and this would override that portion of the prefix.

cvprac/cvp_client.py
cheynearista: Mmm..not sure. I see why you did it (at least I think I do) based on the requirements for requests connect.
```
        :param verify: (optional) Either a boolean, in which case it controls whether we verify
            the server's TLS certificate, or a string, in which case it must be a path
            to a CA bundle to use. Defaults to ``True``.
```

cvprac/cvp_client.py
cheynearista: sure..but you could add self.port and reduce/simplify your code a couple lines. Or am I missing something?

mharista: Nope, you're right. I'll update it.

cvprac/cvp_client.py
mharista: For now I'll add a more detailed parameter description.

PR #41, opened 2017-10-23
Fix get_device_by_name to return only the device with the given FQDN

targuan:
get_device_by_name should ensure that the FQDN of the returned device fully match the parameter.

Before this PR, if a device FQDN is contained in another device FQDN it can't be assured which one is returned.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -275,7 +275,12 @@ class CvpApi(object):
                              'queryparam=%s&startIndex=0&endIndex=0' % fqdn,
                              timeout=self.request_timeout)
         if len(data['netElementList']) > 0:
-            device = data['netElementList'][0]
+            for netElement in data['netElementList']:
+                if netElement['fqdn'] == fqdn:
+                    device = netElement
+                    break
+            else:
+                device = {}
         else:
             device = {}
         return device
--- a/test/system/test_cvp_api.py
+++ b/test/system/test_cvp_api.py
@@ -318,6 +318,13 @@ class TestCvpClient(DutSystemTest):
         self.assertIsNotNone(result)
         self.assertEqual(result, {})
 
+    def test_api_get_device_by_name_substring(self):
+        ''' Verify get_device_by_name with partial fqdn returns nothing
+        '''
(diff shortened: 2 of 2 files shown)


cheynearista:
Interesting case...trying to figure out how you can have the same FQDN for two devices? I could see potentially hitting this issue if you provided a substring that had commonality between multiple FQDNs. For example, you have devices node1.domain.com, node2.domain.com, node3.domain.com, and you specified 'node' as the FQDN to get_device_by_name.

-c

targuan:
One could imagine a naming convention were some FQDN may be included in another FQDN.
For instance if you have something like role.site.domain.com and you have leaf and borderleaf as role, when searching for leaf.site.domain.com you can end up with borderleaf.site.domain.com.

It is a corner case ... but I think the api should be consitent between get_container_by_name where there is a check that the name is actually the one passed as argument and get_device_by_name where anything can be returned as long as the FQDN contains what is passed as argument.

IMHO the device returned by get_device_by_name should be a full match to the argument or the method should be renamed search_device_by_name and return a list.

cheynearista:
@targuan Totally agree. Thanks for the reply. I was more fishing for an additional example case for unittest. You provided just what we needed :)

devrobo:
Would it be possible to get a test case for this change?

targuan:
I'm not used to nose framework for testing, but I added a test case (I hope in the right place) for this PR.

cvprac/cvp_api.py
```diff
@@ -275,7 +275,12 @@ def get_device_by_name(self, fqdn):
                              'queryparam=%s&startIndex=0&endIndex=0' % fqdn,
                              timeout=self.request_timeout)
         if len(data['netElementList']) > 0:
-            device = data['netElementList'][0]
+            for netElement in data['netElementList']:
+                if netElement['fqdn'] == fqdn:
```
mharista: If the call returns several devices that include the provided fqdn but none of them are a perfect match then device will not be set to an empty dict.

mharista: Ignore this comment. I mistakenly assumed for/else clause behaved in the opposite manner that it actually does.

PR #52, opened 2018-04-04
Remove 'id': 1 from _add_temp_action data struct

### LINKED ISSUE #50: deploy_device breaks in 2018.1
looks like we need to use different API endpoints.  Found this out while helping out the SBUX POC.

grybak-arista:
The `'id': 1` key/value included in the data dictionaries passed to _add_temp_action causes CVP 2018 to give an error like
```
CvpApiError: POST: https://10.16.129.30:443/web/provisioning/addTempAction.do?format=topology&queryParam=&nodeId=root : Request Error: Input json is not compliant with schema. object instance has properties which are not allowed by the schema: ["id"]
```

It looks like this key is not required in 2017 versions either, but does not cause the failure seen in 2018.

I think this fixes #50, since deploy_device calls several methods that included this key/value in the data structure.

Testing was only done with cloud-deploy deployments for both 2017 and 2018.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -506,8 +506,7 @@ class CvpApi(object):
 
         info = '%s: Configlet Assign: to Device %s' % (app_name, dev['fqdn'])
         info_preview = '<b>Configlet Assign:</b> to Device' + dev['fqdn']
-        data = {'data': [{'id': 1,
-                          'info': info,
+        data = {'data': [{'info': info,
                           'infoPreview': info_preview,
                           'note': '',
                           'action': 'associate',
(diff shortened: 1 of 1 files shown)


cheynearista:
@grybak-arista So this doesn't work with 'id' set to 1? I find it hard to believe this fails for 2018.1.0. They have the same value coded in their python CVP API library.

dathelen:
@cheynearista I can confirm this was one of the issues we found in the SBUX POC.  There was two other small changes i had to make too to get deploy_device to work.  Once my contact gets back from FTA I will document those in #50

grybak-arista:
@cheynearista I don't see a consistency either. I looked at /cvpi/tools/cvpService.py and saw some calls to addTempAction where the data structure had 'id' set (always to 1) and some calls without the 'id' key.

The change proposed here is what worked for the deployments I was doing, though. Maybe there's a deeper issue at work, but I did not try to look into it more.

cheynearista:
@grybak-arista @dathelen OK. I was trying to figure out if go-cvprac would also have the same issue. And it might. It appears the CVP python api doesn't use id:1 for deployDevice....but as you said uses it other places. I've sent a note to development on the purpose of 'id'. I'll circle back when I hear from them.

PR #56, opened 2018-04-26
add get_configlets function

brokenpackets:


brokenpackets:
Added function for get_configlets to cvp_api.py

default pulls all configlets, can be filtered down to specific types of configlets, or search strings.

mharista:
Can you add tests associated with this new function?

brokenpackets:
Added test matching some of the other get_{items}. Let me know if you think we need something more specific.

brokenpackets:
Closing out - will review current dev branch later and submit fresh PR if still required.

PR #57, opened 2018-04-29
Add capture_ContainerLevelSnapshot Function

brokenpackets:
Add function to leverage snapshot/captureContainerLevelSnapshot.do? api.


mharista:
This PR appears to overlap with code in other open PRs. For example the get_EventDatabyId function is also included here. Please remove duplicate code from this PR and have it specific to the capture_ContainerLevelSnapshot function. Also can you add tests associated with the new function?

PR #58, opened 2018-04-29
Add Function: get_EventDatabyId

brokenpackets:
Add function to query CVP for parent eventID, and returns status of child events.
    Leverages /event/getEventDataById.do api.


mharista:
Can you add tests associated with this new function?

brokenpackets:
Need to rework this in a clean branch. Closing, will re-open.

PR #59, opened 2018-05-05
Issue 50

brokenpackets:
Add option to define if create_task is required when calling the function - fixes issues with for loops in 2018.1.


mharista:
Is this full PR description still accurate? I don't see any updates to configlet changing code.

brokenpackets:
Thanks @mharista - updated the PR description to reflect that this is only adding the option to create a task instead of doing so implicitly. The configlet assign behavior in 2018.1.0 (I believe, could be earlier, been a bit since I've looked at it) for provisioning made it so that once a task is created on a device in the undefined container, configlets can no longer be applied. Requires that user perform all required actions before creating task.

mharista:
Gotcha. Thanks @brokenpackets!

PR #64, opened 2018-07-11
Create import_certificate function

brokenpackets:
Import certificates into CVP and automatically applies the config. Built a test that imports a "fake" cert/key, but keep in mind that by nature this can make CVP stop responding via browser until page is hard-refreshed or window closed/opened.


brokenpackets:
Closing this out, haven't tested any of the api changes to see if it's still relevant.

PR #69, opened 2018-12-04
adding support for urllib and python3 print function

packetferret:
urllib now nests the quote_plus method inside of urllib.parse, adding a try/expect loop for instances where either python2 or 3 are in operation.

also adding in python3's print_function, helps with the whole unicode thing ;)

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -252,9 +252,14 @@ class CvpApi(object):
                 configlet (dict): The configlet dict.
         '''
         self.log.debug('get_configlets_by_name: name: %s' % name)
-        return self.clnt.get('/configlet/getConfigletByName.do?name=%s'
-                             % urllib.quote_plus(name),
-                             timeout=self.request_timeout)
+        try:
+            return self.clnt.get('/configlet/getConfigletByName.do?name=%s'
+                                % urllib.quote_plus(name),
+                                timeout=self.request_timeout)
+        except AttributeError:
+            return self.clnt.get('/configlet/getConfigletByName.do?name=%s'
+                                % urllib.parse.quote_plus(name),
+                                timeout=self.request_timeout)
 
     def get_configlets_by_container_id(self, c_id, start=0, end=0):
         ''' Returns a list of configlets applied to the given container.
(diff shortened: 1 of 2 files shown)


cvprac/cvp_client.py
```diff
@@ -193,23 +194,23 @@ def set_version(self, version):
...
                 self.apiversion = 'v2'
             else:
-                print 'FOUND V1'
+                print('FOUND V1')
                 self.apiversion = 'v1'
         else:
             # If version is shorter than 2 elements for some reason default
             # to v2
-            print 'SHORT VERSION STRING SET v2'
+            print('SHORT VERSION STRING SET v2')
```
mharista: Could you revert changes to cvp_client.py? I've converted these print statements to log.info statements in the latest release and develop branch.

cvprac/cvp_api.py
```diff
@@ -667,9 +687,14 @@ def add_configlet(self, name, config):
                        timeout=self.request_timeout)
 
         # Get the key for the configlet
-        data = self.clnt.get('/configlet/getConfigletByName.do?name=%s'
-                             % urllib.quote_plus(name),
-                             timeout=self.request_timeout)
+        try:
```
mharista: I am going to think about this try/except format for a day. Functionality looks good but we may be able to handle this in a cleaner way.

packetferret: Agreed, there is likely a much cleaner way of performing this. I had considered created a new object and setting it to the method of urllib.quote_plus or urllib.parse.quote_plus based on the version of python the user is running.

looking forward to seeing what you come up with

cvprac/cvp_client.py
packetferret: I certainly can, but the cvprac package's installation fails on my macOS laptop with python3 without the parentheses

mharista: The print statements are removed in the latest release 1.0.0. Can you try installing that one and let me know if that fails? The only remaining prints are the ones in the docs.

cvprac/cvp_api.py
mharista: Can you try removing the try/except clauses in the methods and apply it during the imports?

try:
    from urllib import quote_plus as qplus
except AttributeError:
    from urllib.parse import quote_plus as qplus

And then in the methods you always reference the alias qplus.

Example:
return self.clnt.get('/configlet/getConfigletByName.do?name=%s'
                             % **qplus(name)**, timeout=self.request_timeout)

packetferret:
addressing in #73

PR #82, opened 2019-02-01
Poll for inventory in v2 add_device_to_inventory

grybak-arista:
It can take a moment for device to appear. This polls the inventory for 60 seconds to see if the device is added.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -362,6 +362,48 @@ class CvpApi(object):
                     dev['containerName'] = ''
             return data
 
+    def add_devices_to_undefined(self, device_ips, wait=True):
+        ''' Add the list of devices to the undefined container.
+
+            Args:
+                device_ips (list): ip addresses of devices to be added
+                wait (boolean): If True, wait up to 60 seconds for
+                    new device to appear in the undefined container
+        '''
+
+        self.log.debug('add_devices_to_undefined: called')
+        if self.clnt.apiversion is None:
+            self.get_cvp_info()
+        if self.clnt.apiversion == 'v1':
+            self.log.debug('v1 Inventory API Call')
+            self.log.warning(
+                'add_devices_to_undefined: not implemented for v1 APIs')
+        else:
+            self.log.debug('v2 Inventory API Call')
+            data = {'hosts': device_ips}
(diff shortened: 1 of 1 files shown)


cvprac/cvp_api.py
```diff
@@ -361,6 +362,48 @@ def get_inventory(self, start=0, end=0, query=''):
...
+
+            if wait:
+                # It can take a few moments for devices to appear in inventory
+                dev = None
+                timeout = time.time() + 60      # Try for 60 seconds
+                while not dev and time.time() < timeout:
+                    devices = self.get_inventory()
+                    for device in devices:
+                        if ('ipAddress' in device and
+                                device['ipAddress'] == device_ip):
```
mharista: Does this work? Where is device_ip defined?

cvprac/cvp_api.py
```diff
@@ -361,6 +362,48 @@ def get_inventory(self, start=0, end=0, query=''):
...
+                            dev = device
+                            break
+                    if not dev:
+                        time.sleep(5)
+
+                if not dev:
+                    # Device did not show up in inventory
+                    raise RuntimeError('Device {} failed to appear in '
+                                       'undefined inventory'
+                                       .format(device_ip))
```
mharista: Where is device_ip defined?

cvprac/cvp_api.py
```diff
@@ -361,6 +362,48 @@ def get_inventory(self, start=0, end=0, query=''):
...
+                'add_devices_to_undefined: not implemented for v1 APIs')
+        else:
+            self.log.debug('v2 Inventory API Call')
+            data = {'hosts': device_ips}
+            self.clnt.post('/inventory/devices', data=data,
+                           timeout=self.request_timeout)
+
+            if wait:
+                # It can take a few moments for devices to appear in inventory
+                dev = None
```
mharista: This function take multiple device addresses to be added to undefined but only checks that one of them has been added during the wait?

devrobo:
Are we sure we want the library to poll?

Gary,
Do you need this in the test environment because the topology just spun up and you are waiting for the devices to appear? This shouldn't be needed during normal operation correct?

grybak-arista:
Updated changes and submitted as a new pull request.

PR #84, opened 2019-04-20
Add multiple devices in one call

grybak-arista:
Add function add_devices_to_inventory to allow multiple devices to be added simultaneously. Helps speed up large deployments from cloud-deploy. Basically a multiple device version of the existing add_device_to_inventory.

In the v2 api, the call to inventory/devices sometimes takes a short while for the devices to actually appear in the inventory list. We need to wait for the devices to appear before we can move them to their respective containers.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -32,6 +32,7 @@
 ''' Class containing calls to CVP RESTful API.
 '''
 import os
+import time
 # This import is for proper file IO handling support for both Python 2 and 3
 # pylint: disable=redefined-builtin
 from io import open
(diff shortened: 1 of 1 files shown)


cvprac/cvp_api.py
```diff
@@ -361,6 +362,86 @@ def get_inventory(self, start=0, end=0, query=''):
...
+                data_list.append(dev_data)
+            data = {'data': data_list}
+            self.clnt.post('/inventory/add/addToInventory.do?'
+                           'startIndex=0&endIndex=0', data=data,
+                           timeout=self.request_timeout)
+        else:
+            self.log.debug('v2 Inventory API Call')
+
+            # First add the devices to the inventory in a single call
+            data = {'hosts': [x['device_ip'] for x in device_list]}
```
mharista: Consolidate list creation of device_ips into one line up above this and remove 420.

cvprac/cvp_api.py
```diff
@@ -361,6 +362,86 @@ def get_inventory(self, start=0, end=0, query=''):
...
+            self.clnt.post('/inventory/devices', data=data,
+                           timeout=self.request_timeout)
+
+            # With v2, the devices can take a few moments to appear
+            # We need them present before we can move them to a container
+            device_ips = [x['device_ip'] for x in device_list]
```
mharista: Consolidate list creation of device_ips into one line before 414. Remove this one.

cvprac/cvp_api.py
```diff
@@ -361,6 +362,86 @@ def get_inventory(self, start=0, end=0, query=''):
...
+            inv = self.get_inventory()
+            timeout = time.time() + 600
```
mharista: In general if there is a potential timeout issue for v2 between adding of a device to inventory and moving it to its parent container shouldn't this timeout be added or optional with the single device version of this method? Is this only an issue when multiple devices are added to inventory in a single call?

cvprac/cvp_api.py
```diff
@@ -361,6 +362,86 @@ def get_inventory(self, start=0, end=0, query=''):
...
+            while device_ips and time.time() < timeout:
+                inv_devices = [x['ipAddress'] for x in inv]
```
mharista: Is using 'x' in this syntax standard for Python? I personally think using 'dev' in the place of 'x' for all of these lines is more readable.

cvprac/cvp_api.py
```diff
@@ -361,6 +362,86 @@ def get_inventory(self, start=0, end=0, query=''):
...
+            # Move the devices to their specified containers
+            for device in device_list:
+                devs = [x for x in inv if 'ipAddress' in x and
+                        device['device_ip'] in x['ipAddress']]
+                dev = devs[0]
+                container = {'key': device['parent_key'],
+                             'name': device['parent_name']}
+                self.move_device_to_container('add_device_to_inventory API v2',
+                                              dev, container, False)
+
```
mharista: Need a systest for this function.

cvprac/cvp_api.py
grybak-arista: Yes, the timeout should be optional for both single and multiple device versions. Meant to add that in as a 'wait' option, defaulting to False.

Also thinking maybe single device version should just call multiple device version with a one element list. Does that make sense?

cvprac/cvp_api.py
grybak-arista: Single letter variables for list comprehensions seems to be commonly used, my guess is because the comprehension can become very long when the tests are more complex. I'll change this and the device_ips to use 'dev' for clarity though.

cvprac/cvp_api.py
grybak-arista: Will do

cvprac/cvp_api.py
mharista: Ya I think that makes sense and would allow the multiple device function to leverage the current single device function systest.

cvprac/cvp_api.py
mharista: That's a good point. Maybe keep it a single letter but go with 'd' instead of 'x'?

cvprac/cvp_api.py
mharista: I think we could leverage the single device systest if we consolidate the single device function and multi device function as you mentioned above.

grybak-arista:
@mharista Updated for the requested changes.

Did not add a specific test case for multiple device addition, as suggested since single device now calls the multiple device version with a one-device list. Specific test case can be added if needed, but will change the test suite requirements to contain more than just one device.

PR #95, opened 2019-07-24
Add search_configlets method and update version test

grybak-arista:
search_configlets: calls configlet/searchConfiglets.do and returns the result.

Update the set_version method to set the api version to v1 or v2 based on CVP version used. Previous method split the version string and compared each element, so 2019.1 elements were not *both* >2017 and >1, so it applied api v1. Use pkg_resources.parse_version instead, which handles minor versions and even -EFT tags well.


grybak-arista:
This got pushed aside and forgotten. Issue #99 reminded me. Fairly certain I tested against multiple versions, but I will run tests against various v1 and v2 instances again just to reassure myself.

Issue #103, opened 2020-05-21
cvprac is using the incorrect version of the API for 2018.1.x


colinmacgiolla:
When attempting to use cvprac against CVP 2018.1.3 it fails because it doesn't distinguish between the train and the running version correctly.

`parse_version("2018.1.3") >  parse_version('2018.1')
True
`
Will have cvprac using v2 of the APIs rather then v1

PR #104, opened 2020-05-21
CVPRAC using the wrong API version

colinmacgiolla:
CVPRAC is selecting the wrong API version as 2018.1.3 > 2018.1


colinmacgiolla:
Fixes #103

mharista:
Thanks for fixing this Colin.

PR #113, opened 2020-07-20
adding update_username function to change local username password

noredistribution:


cvprac/cvp_api.py
```diff
@@ -96,6 +96,26 @@ def get_cvp_info(self):
             self.clnt.set_version(data['version'])
         return data
 
+    def update_username(self, username, password, role):
```
mharista, line 99: The function name update_username seems a bit misleading. Isn't this function technically updating a given user accounts password and role? It isn't actually changing the username itself is it? If my assumption is right maybe the function name of update_user, or update_user_info, or update_user_data?

noredistribution: yes you're right, we could go for `update_user` which is close to the name of the API (`updateUser.do`) or `update_user_info` ?

mharista: I like update_user as it keeps the name shorter and closer to the API

cvprac/cvp_api.py
```diff
@@ -1,4 +1,4 @@
-#
+#F
```
mharista, line 1: I'll remove this extra 'F' character.

test/system/test_cvp_api.py
```diff
@@ -199,6 +199,34 @@ def test_api_get_cvp_info(self):
         self.assertEqual(self.clnt.version, result['version'])
         self.assertIsNotNone(self.clnt.apiversion)
 
+    def test_api_get_user(self):
```
mharista, line 202: I will consolidate these tests into one function.

PR #123, opened 2020-09-30
Never fallback to HTTP in case of connection failure

freedge:
Current code attempts a connection with https, and upon failure (eg: a mistake in the password) falls back to http. As a consequence:
- the initial error (why it failed in the first place) is lost. It makes the investigation process much difficult
- the password is sent unencrypted over the network.

Apparently the initial behavior of "defaulting to http" was a request for users of old CVP versions that were fine configuring their network devices over an unencrypted connection.

Fix it: remove this feature entirely.


mharista:
Hi @freedge 

Thanks for adding this. Going to have a brief internal discussion about this before merging.

mharista:
Thanks for this addition @freedge

PR #128, opened 2020-12-03
started to add api method update_configlet_builder

dbm79:
This is the first time I have made any meaningful contribution to open source so I apologize if I am doing something wrong.
 
I started to add a new API method in order to update configlet builders called update_configlet_builder.  I noticed this was lacking and so far I have been manually doing it with CVP's API in my own scripts..  Simply doing an updateConfiglet API would cause it to be of type static and not builder.  I hope this is helpful

I haven't tested this yet as far as this project goes.  Please feel free to reach out to me or let me know if i messed anything up.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -1092,6 +1092,32 @@ class CvpApi(object):
         return self.clnt.post('/configlet/updateConfiglet.do', data=body,
                               timeout=self.request_timeout)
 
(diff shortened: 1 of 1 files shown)


mharista:
Hi @dbm79 

Thanks for the addition. In general when adding new functions we also add associated test functions. In your case it may not be as easy to create the appropriate test environment so I will look into merging this and adding the test functions.

PR #134, opened 2021-01-25
Provide timestamp by default for approving CC

colinmacgiolla:
Suggested enhancement for the approve_change_control

Basically the change provides the timestamp by default, which makes the use of this call more intuitive, while not breaking any existing users

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,7 +15,11 @@ Table of Contents
 
 #. `Getting Started`_
 
-   -  `Example`_
+   -  `Connecting`_
+   -  `CVP On Premises`_
+   -  `CVaaS`_
+   -  `CVP Version Handling`_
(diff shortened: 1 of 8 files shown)


mharista:
I like the idea at first look but I'm going to think for a bit about whether or not we want this to be the default behavior

colinmacgiolla:
Yeah, I mean for me it felt really weird that the when I was calling the API I needed to provide the timestamp, but the logic makes sense from a wider context, where we are basically saying I approve this change at this time.

My thinking was that by providing the new TS by default, we make it cleaner for new users, but at the same time don't break it for anyone that has already integrated cvprac

mharista:
I agree, I think this is good/intuitive default behavior.

Issue #141, opened 2021-03-25
Update client to handle decoding multiple json objects


mharista:
>>> cvpevents = client.get('/api/resources/event/v1/Event/all')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/site-packages/cvprac/cvp_client.py", line 858, in get
    return self._make_request('GET', url, timeout)
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/site-packages/cvprac/cvp_client.py", line 641, in _make_request
    response = self._send_request(req_type, full_url, timeout,
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/site-packages/cvprac/cvp_client.py", line 785, in _send_request
    self._is_good_response(response, '%s: %s ' %
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/site-packages/cvprac/cvp_client.py", line 409, in _is_good_response
    joutput = response.json()
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/site-packages/requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/Users/tamas/.pyenv/versions/3.8.6/lib/python3.8/json/decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 2 column 1 (char 561)

Data is returned in format similar to below:

{"key":value}
{"key":value}

mharista:
Fixed #142

PR #142, opened 2021-04-06
Add client handling for new resource API REST bindings that return multiple objects in response data

mharista:


mharista:
Fixes #141

Issue #144, opened 2021-04-30
delete_container() return 'success' even when it isn't successful


choy11:
Removing an empty container works as expected.  When ran against a non-empty container, I still get "{'data': {'taskIds': [], 'status': 'success'}}" returned and the container remains in CVP.  I expected to get an error back saying the container wasn't empty but that wasn't that case.

mharista:
Hi @choy11 

I've been looking into this and have found that with the current REST APIs it is very inefficient to check if a container being deleted has any child devices or containers.

Based on this I am planning to run a check after the attempted delete to read the container that was just deleted. If it is found after the delete I will return a status of 'fail' with brief details or I will raise an exception.

mharista:
Interestingly this seems to have changed somewhere in CVP. I just found that CVP 2018.2.5 raises an exception when trying to delete a container with children. Will try to narrow down where this changed.

choy11:
"Based on this I am planning to run a check after the attempted delete to read the container that was just deleted. If it is found after the delete I will return a status of 'fail' with brief details or I will raise an exception."

That is exactly what I did in my script and is acceptable if there is no other way.  Thank you for looking into it!

mharista:
Error from 2018.2.5

CvpApiError: POST: https://10.16.129.153:443/web/provisioning/addTempAction.do?format=topology&queryParam=&nodeId=root : Request Error: Only empty container can be deleted

mharista:
Hi @choy11 

Now that I've found that this has changed at some point in CVP I'm curious what version of CVP you are running?

choy11:
2020.3.1

choy11:
Sorry, that was the wrong window, I am working in two different instances right now.  The one I am running the scripts against is 2020.3.0

mharista:
@choy11 

Thanks for the update. I've found that the delete_container function stops raising an error from addTempAction.do when the container has children in CVP 2020.1.2. 

Given that the prior functionality was to raise an exception I will repeat that functionality going forward. Essentially if I find the container still exists after the delete I will raise a CvpApiError as CVP did prior to CVP 2020.1.2.

mharista:
Fixed #145

choy11:
You are awesome, thank you!

PR #145, opened 2021-05-14
Add exception when attempting to delete container with children for CVP versions 2020.1 and beyond

mharista:


mharista:
Fixes #144

Issue #152, opened 2021-06-04
Add support for search device by hostname


titom73:
For ansible-cvp collection, we are moving to a search by hotname instead of FQDN as it can create issue when we provision devices with no `dns domain` and then apply such configuration.

Currently, we don't have such function in cvprac and we can only rely on `def get_device_by_name(self, fqdn)` which requires an FQDN to get device.

Idea would be to do one of these options:

- Create a new function to allow search by FQDN field
- Implement an option in `def get_device_by_name(self, fqdn)` to select field to use for search: `fqdn` or `hostname`

Happy to discuss about the best approach before sending PR.

titom73:
Actually, `hostname` is not part of the response when running search_topology

```json
{
  "keywordList": [],
  "total": 1,
  "netElementList": [
    {
      "modelName": "",
      "internalVersion": "4.25.2F",
      "systemMacAddress": "50:8d:00:e3:78:aa",
      "serialNumber": "79AEA53101E7340AEC9AA4819D5E1F5B",
      "memTotal": 0,
      "bootupTimeStamp": 1622794376.503242,
      "memFree": 0,
      "version": "4.25.2F",
      "architecture": "",
      "internalBuildId": "6abe91ca-cb14-4038-96f6-f83dd0a346f5",
      "hardwareRevision": "",
      "fqdn": "CV-ANSIBLE-EOS01.emea.lab",
      "key": "50:8d:00:e3:78:aa"
```

So, having option to compare only hostname part of the FQDN would be nice

emilotep:
Have a customer that needs this in order to move to using v3 cvp modules with AVD.

mharista:
Hi @emilotep 

I'll look into getting this merged today

mharista:
Fixed #153

PR #153, opened 2021-06-04
feat(cvp_api): Add support for hostname search in get_device_by_name

titom73:
Allow to search device by using hostname part of the FQDN and not the full FQDN field.

Tentative fix for #152 

```python
# Default search using FQDN
device_facts = self.__cv_client.api.get_device_by_name(fqdn=device.fqdn)

# Force to use FQDN to search
device_facts = self.__cv_client.api.get_device_by_name(fqdn=device.fqdn, search_field='fqdn')

# Force to use hostname part of FQDN to search
device_facts = self.__cv_client.api.get_device_by_name(fqdn=device.fqdn, search_field='hostname')
```

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -810,11 +810,12 @@ class CvpApi(object):
                     devices.append(device)
         return devices
 
-    def get_device_by_name(self, fqdn):
+    def get_device_by_name(self, fqdn, search_by_hostname=False):
         ''' Returns the net element device dict for the devices fqdn name.
 
             Args:
-                fqdn (str): Fully qualified domain name of the device.
+                fqdn (str): Fully qualified domain name or hostname of the device.
+                search_field (str): method to search for device: fqdn or hostname.
 
             Returns:
                 device (dict): The net element device dict for the device if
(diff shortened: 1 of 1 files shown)


mharista:
Hi @titom73 

I like this idea overall but am going to think about some minor tweaks. For example I'm thinking about changing the new parameter from "search_field" because as you pointed out in the issue there is no hostname field in the response so having search_field set to hostname is a bit misleading. I'm thinking about maybe a simple flag to tell the function to parse the fqdn for a matching hostname.

mharista:
Hi @titom73 

Are you opposed to updating the new parameter to a boolean defaulting to False named search_by_hostname? If set True we simply do the 'fqdn' to provided name comparison based on the first index of the split on the 'fqdn' index as you are currently doing.

titom73:
Hello @mharista 

Thanks for the feedback. I am fine changing this new parameter to use a boolean one instead of string. Let me change the code to move to this approach.

I will update this PR over the week-end.

mharista:
Hi @titom73 

I've seen some additional requests for this on the issue so I am planning to merge your code and make the updates I've suggested today.

PR #167, opened 2021-09-02
Some enhancements to the task_retrigger script

colinmacgiolla:
Added some things (listed below) to make the script more general purpose;
- Accepts arguments for username, password and cvp servers
- Handles unverified context for older python versions
- Triggers a compliance check at the beginning (may require a delay?)
- Won't attempt task generation for unregistered devices or devices in the undefined container
- Allows for string match filtering of the devices its going to try and update

### DIFF
--- a/docs/labs/lab06-provisioning/vc_task_retrigger.py
+++ b/docs/labs/lab06-provisioning/vc_task_retrigger.py
@@ -116,3 +116,4 @@ def main():
 
 if __name__ == "__main__":
     main()
+    
--- a/docs/labs/lab06-provisioning/vc_task_retrigger.py
+++ b/docs/labs/lab06-provisioning/vc_task_retrigger.py
@@ -7,6 +7,7 @@
 import argparse
 import ssl
 import sys
+from packaging import version
 from getpass import getpass
 from cvprac.cvp_client import CvpClient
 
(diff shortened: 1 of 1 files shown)


docs/labs/lab06-provisioning/vc_task_retrigger.py
```diff
@@ -4,39 +4,109 @@
...
+        print("Connecting to %s" % cvpserver)
+        try:
+            clnt.connect(nodes=[cvpserver], username=args.username, password=args.password)
+        except Exception as e:
+            print("Unable to connect to CVP: %s" % str(e))
+        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+        # Trigger compliance check here
+        #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!1
+        try:
+            clnt.api.check_compliance('root', 'container')
```
noredistribution, line 62: Maybe we should add a note here? as we don't need to do this post 2020.3.0 where we have continous compliance check, so running `get_device_by_name(device)` or `get_inventory()` will get you always the latest compliance codes

noredistribution (commented):
thanks for the PR, LGTM, left a comment

docs/labs/lab06-provisioning/vc_task_retrigger.py
colinmacgiolla: Or just add a check - only do this if the CVP release is <2020.3.0

noredistribution: good idea! you could use `get_cvp_info()`

mharista: @colinmacgiolla are you planning to add the version check before I merge this?

colinmacgiolla:
Yes, I'll add it over the weekend

mharista (changes requested):
Github is notifying that the file is missing a newline at the end. Can you add this and then I think we are good to merge.

colinmacgiolla:
Done

PR #168, opened 2021-09-04
api: add workspace and tagv2 resource apis

noredistribution:
Phase1: 

Added the following APIs:

tag.v2.TagConfig - create/remove device or interface tags
tag.v2.TagConfig/all - show all tags edits in a workspace
tag.v2.TagAssignmentConfig - Assign/Unassign device or interface tags
tag.v2.TagAssignmentConfig/all - set all tag assignment edits within a workspace
tag.v2.Tag/all - get all tags
workspace.v1.WorkspaceBuild - verify build status
workspace.v1.WorkspaceConfig - create/abandon/rollback/submit workspace and start/cancel build


mharista (commented):
I've been thinking about creating a new API class for Resources to break them away from the standard API instead of continuing to pile onto the existing (and huge) API file. My understanding is that in the long run the resource APIs will fully replace the old REST apis and hence we would eventually be able to deprecate the old API class and file.

PR #175, opened 2021-11-19
get the running-config of devices at specified time

noredistribution:
Added simple example on how to use the compliance service API to get the RC of each device in the inventory at a specific time and write it to a file


guillaumeVilar (approved):
LGTM, tested and it created a .cfg file per device - the devices not present at that time are printing an error message as expected

mharista (approved):

PR #179, opened 2021-12-17
Add cc rapi

noredistribution:
This PR is to add wrappers for the new [Change Control Resource APIs ](https://aristanetworks.github.io/cloudvision-apis/models/changecontrol.v1/)introduced in 2021.2.0

The following new functions were added:
- change_control_get_one
- change_controls_get_all
- change_control_approval_get_one
- change_control_approval_get_all
- change_control_approve
- change_control_delete
- change_control_create_with_custom_stages
- change_control_create_for_tasks
- change_control_start
- change_control_stop
- change_control_schedule

All functions have detailed docstrings
CC creation is possible either via `change_control_create_for_tasks` which can generate a CC based on a list of task and run them in series or parallel, or more complex CCs can be created with `change_control_create_with_custom_stages` where the dictionary that holds the stage hierarchy has to be provided by the user (two examples provided in the docstring)

Here are two examples (we'll add these to the lab section later):
Ex1) Create simple CC

```python
from cvprac.cvp_client import CvpClient
import ssl
import uuid
from datetime import datetime
ssl._create_default_https_context = ssl._create_unverified_context
import requests.packages.urllib3
requests.packages.urllib3.disable_warnings()
clnt = CvpClient()
clnt.connect(['cvp'],'username', 'password')

cc_id = str(uuid.uuid4())
name = f"Change_{datetime.now().strftime('%Y%m%d_%H%M%S')}"

# Create CC with all tasks to be executed in parallel
tasks = ["1240","1241","1242","1243"]
print(clnt.api.change_control_create_for_tasks(cc_id, name, tasks, series=True)
```

Ex2) Create complex CC with custom hierarchy

```python
from cvprac.cvp_client import CvpClient
import ssl
import uuid
from datetime import datetime
ssl._create_default_https_context = ssl._create_unverified_context
import requests.packages.urllib3
requests.packages.urllib3.disable_warnings()
clnt = CvpClient()
clnt.connect(['cvp'],'username', 'password')
cc_id = str(uuid.uuid4())
name = f"Change_{datetime.now().strftime('%Y%m%d_%H%M%S')}"

data = {'key': {
            'id': cc_id
            },
        'change': {
            'name': cc_id,
            'notes': 'cvprac CC',
            'rootStageId': 'root',
            'stages': {'values': {'root': {'name': 'root',
                                           'rows': {'values': [{'values': ['1-2']},
                                                               {'values': ['3']}]
                                            }
                                        },
                                  '1-2': {'name': 'stages 1-2',
                                          'rows': {'values': [{'values': ['1ab']},
                                                              {'values': ['2']}]}},
                                  '1ab': {'name': 'stage 1ab',
                                          'rows': {'values': [{'values': ['1a','1b']}]
                                            }
                                        },
                                  '1a': {'action': {'args': {'values': {'TaskID': '1242'}},
                                                    'name': 'task',
                                                    'timeout': 0},
                                         'name': 'stage 1a'},
                                  '1b': {'action': {'args': {'values': {'TaskID': '1243'}},
                                                    'name': 'task',
                                                    'timeout': 0},
                                         'name': 'stage 1b'},
                                  '2': {'action': {'args': {'values': {'TaskID': '1240'}},
                                                   'name': 'task',
                                                   'timeout': 0},
                                        'name': 'stage 2'},
                                  '3': {'action': {'args': {'values': {'TaskID': '1241'}},
                                                   'name': 'task',
                                                   'timeout': 0},
                                        'name': 'stage 3'},
                }
            }
        }
 }

print(clnt.api.change_control_create_with_custom_stages(data))
```

The above would result in the following hierarchy:

 ```
root (series)
 |- stages 1-2 (series)
 |  |- stage 1ab (parallel)
 |  |    |- stage 1a
 |  |    |- stage 1b
 |  |- stage 2
 |- stage 3
```

Added two examples in lab06

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -2943,8 +2943,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('tag_assignment_config: Tag.V2 Resource APIs are supported'
-                                 ' from 2021.2.0 or newer.')
+                self.log.warning('Tag.V2 Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(tag_url))
         return self.clnt.post(tag_url, data=payload)
@@ -2975,9 +2974,10 @@ class CvpApi(object):
         if not self.clnt.is_cvaas:
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
-            if self.clnt.apiversion >= 6.0:
-                self.log.debug('v6 ' + tag_url + ' ' + str(payload))
-                return self.clnt.post(tag_url, data=payload)
+            if self.clnt.apiversion < 6.0:
+                self.log.warning('Tag.V2 Resource APIs are supported from 2021.2.0 or newer.')
+                return None
+        self.log.debug('v6 ' + tag_url + ' ' + str(payload))
         return self.clnt.post(tag_url, data=payload)
 
     def get_tag_assignment_edits(self, workspace_id):
@@ -3007,8 +3007,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('tag_assignment_config: Tag.V2 Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Tag.V2 Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 ' + tag_url + ' ' + str(payload))
         return self.clnt.post(tag_url, data=payload)
@@ -3048,8 +3047,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('tag_assignment_config: Tag.V2 Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Tag.V2 Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {} '.format(tag_url) + str(payload))
         return self.clnt.post(tag_url, data=payload)
@@ -3097,8 +3095,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('tag_assignment_config: Tag.V2 Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Tag.V2 Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {} '.format(tag_url) + str(payload))
         return self.clnt.post(tag_url, data=payload)
@@ -3116,8 +3113,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                    self.log.warning('get_all_workspaces: Workspace Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                    self.log.warning('Workspace Resource APIs are supported from 2021.2.0 or newer.')
                     return None
         self.log.debug('v6 {}'.format(url))
         return self.clnt.post(url, data=payload)
@@ -3134,8 +3130,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('get_workspace: Workspace Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Workspace Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url))
         return self.clnt.get(url)
@@ -3186,8 +3181,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('workspace_config: Workspace Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Workspace Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 ' + str(url) + ' ' + str(payload))
         return self.clnt.post(url, data=payload)
@@ -3211,8 +3205,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('workspace_build_status: Workspace Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Workspace Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url+params))
         return self.clnt.get(url, timeout=self.request_timeout)
@@ -3247,11 +3240,17 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_get_one: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url))
-        return self.clnt.get(url, timeout=self.request_timeout)
+        try:
+            response = self.clnt.get(url, timeout=self.request_timeout)
+        except Exception as error:
+            if 'resource not found' in str(error):
+                self.log.error('Change Control ID does not exist!')
+                return None
+            raise error
+        return response
 
 
     def change_control_get_all(self):
@@ -3267,8 +3266,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_get_all: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url))
         return self.clnt.get(url, timeout=self.request_timeout)
@@ -3298,20 +3296,16 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_approval_get_one: Change Control Resource APIs are supported'
-                                    ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
-        try:
-            cc_status = self.change_control_get_one(cc_id)
-        except Exception as e:
-            self.log.warning('change_control_get_one: The change control ID does not exist.')
+        cc_status = self.change_control_get_one(cc_id)
+        if cc_status is None:
             return None
         if 'approve' not in cc_status['value']:
-            self.log.warning("change_control_approval_get_one: The change has not been approved yet."
-                                " A change has to be approved at least once for the 'approve' state to be populated.")
+            self.log.warning("The change has not been approved yet."
+                             " A change has to be approved at least once for the 'approve' state to be populated.")
             return None
         return self.clnt.get(url, timeout=self.request_timeout)
-
 
 
     def change_control_approval_get_all(self):
@@ -3328,11 +3322,11 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_approval_get_all: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url))
         return self.clnt.get(url, timeout=self.request_timeout)
+
 
     def change_control_approve(self, cc_id, notes="", approve=True):
         ''' Approve/Unapprove a change control using Resource APIs.
@@ -3346,11 +3340,16 @@ class CvpApi(object):
         url = '/api/resources/changecontrol/v1/ApproveConfig'
         # For on-prem check the version as it is only supported from 2021.2.0+
         # Since the get_change_control already checks this, no need to check it again
-        try:
-            version = self.change_control_get_one(cc_id)['value']['change']['time']
-        except Exception as e:
-            version = None
+        # try:
+        #     version = self.change_control_get_one(cc_id)['value']['change']['time']
+        # except Exception as e:
+        #     version = None
+        #     return None
+        cc_status = self.change_control_get_one(cc_id)
+        if cc_status is None:
             return None
+        if 'value' in cc_status and 'change' in cc_status['value'] and 'time' in cc_status['value']['change']:
+            version = cc_status['value']['change']['time']
         payload = {
             "key": {
                 "id": cc_id
@@ -3378,8 +3377,7 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_delete: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 {}'.format(url))
         return self.clnt.delete(url, timeout=self.request_timeout)
@@ -3491,16 +3489,12 @@ class CvpApi(object):
         if not self.clnt.is_cvaas:
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
-            if self.clnt.apiversion >= 6.0:
-                self.log.debug('v6 ' + str(url) + ' ' + str(payload))
-                response = self.clnt.post(url, data=payload)
-                return response
-            elif self.clnt.apiversion < 6.0:
-                    self.log.warning('change_control_create_with_custom_stages: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
-                    return None
-        else:
-            return self.clnt.post(url, data=payload)
+            if self.clnt.apiversion < 6.0:
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
+                return None
+        self.log.debug('v6 ' + str(url) + ' ' + str(payload))
+        return self.clnt.post(url, data=payload)
+
 
     def change_control_create_for_tasks(self, cc_id, name, tasks, series=True):
         ''' Create a simple Change Control for tasks using Resource APIs.
@@ -3548,11 +3542,11 @@ class CvpApi(object):
             if self.clnt.apiversion is None:
                 self.get_cvp_info()
             if self.clnt.apiversion < 6.0:
-                self.log.warning('change_control_create_for_tasks: Change Control Resource APIs are supported'
-                                     ' from 2021.2.0 or newer.')
+                self.log.warning('Change Control Resource APIs are supported from 2021.2.0 or newer.')
                 return None
         self.log.debug('v6 ' + str(url) + ' ' + str(payload))
         return self.clnt.post(url, data=payload, timeout=self.request_timeout)
+
 
     def change_control_start(self, cc_id, notes=""):
         ''' Start a Change Control using Resource APIs.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cvprac/cvp_api.py [hunk 1]
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
...
         return self.clnt.get(url, timeout=self.request_timeout)
+
+
+    def get_change_control(self, cc_id, cc_time=None):
+        ''' Get the configuration and status of a change control using Resource APIs.
+
+            Args:
+               cc_id (str): The ID of the change control.
+               time (str): Time indicates the time for which you are interested in the data.
+                    If no time is given, the server will use the time at which it makes the request.
```
guillaumeVilar, line 3218: doc_string to fix: `time` --> `cc_time`

cvprac/cvp_api.py
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
...
+        return self.clnt.get(url, timeout=self.request_timeout)
+
+
+    def get_change_control_approval(self, cc_id, cc_time=None):
+        ''' Get the state of a specific Change Control's approve config using Resource APIs.
+
+            Args:
+               cc_id (str): The ID of the change control.
+               time (str): Time indicates the time for which you are interested in the data.
+                    If no time is given, the server will use the time at which it makes the request.
```
guillaumeVilar, line 3267: doc_string to fix: `time` --> `cc_time`

guillaumeVilar (commented):
Thanks a lot! Please find my review below.

cvprac/cvp_api.py
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
...
+        else:
+            params = 'key.id={}&time={}'.format(cc_id, cc_time)
+        url = '/api/resources/changecontrol/v1/ChangeControl?' + params
+        # For on-prem check the version as it is only supported from 2021.2.0+
+        if not self.clnt.is_cvaas:
+            if self.clnt.apiversion is None:
+                self.get_cvp_info()
+            if self.clnt.apiversion >= 6.0:
+                return self.clnt.get(url, timeout=self.request_timeout)
+        return self.clnt.get(url, timeout=self.request_timeout)
```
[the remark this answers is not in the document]
noredistribution: yes, so I did that based on the other calls, the idea is that you can only run this against CVaaS or CVP 2021.2.0+ as that's where this particular API is supported. Probably we should add some better error messaging, but perhaps that should be done in the client class in https://github.com/aristanetworks/cvprac/blob/develop/cvprac/cvp_client.py#L196? what do you think @mharista ?

cvprac/cvp_api.py
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
             if self.clnt.apiversion >= 6.0:
                 return self.clnt.get(url, timeout=self.request_timeout)
         return self.clnt.get(url, timeout=self.request_timeout)
+
+
+    def get_change_control(self, cc_id, cc_time=None):
+        ''' Get the configuration and status of a change control using Resource APIs.
+
```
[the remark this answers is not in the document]
noredistribution: so the name of the functions were chosen based on the[ resource API names](https://aristanetworks.github.io/cloudvision-apis/models/changecontrol.v1/) and also in a way to not coincide with the old and deprecated (or soon to be deprecated change control functions) . I mentioned in the docstring for all these if they are using resource API, but maybe we could make it somewhere else clear too, and maybe we could add into the docstring which CVP versions support what CC API, ideally people should read the documentation, I will also add a few labs and soon we will render the website too so hopefully it will be easier to understand what calls to use, what do you think?

cvprac/cvp_api.py
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
...
+            if self.clnt.apiversion >= 6.0:
+                self.log.debug('v6 {}'.format(url))
+                response = self.clnt.get(url, timeout=self.request_timeout)
+                return response
+        return self.clnt.get(url, timeout=self.request_timeout)
+
+
+    def get_change_control_approval(self, cc_id, cc_time=None):
+        ''' Get the state of a specific Change Control's approve config using Resource APIs.
+
```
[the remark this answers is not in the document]
noredistribution: yes, thanks for raising this, this is expected behaviour with the v1 API,  I'll add some error handling to this, basically today if a CC was never approved you cannot get approval status because the key doesn't exist in the db yet (will make this clear in the cloudvision-apis examples doc too)

cvprac/cvp_api.py
```diff
@@ -3207,3 +3207,360 @@ def workspace_build_status(self, workspace_id, build_id):
...
+                self.get_cvp_info()
+            if self.clnt.apiversion >= 6.0:
+                self.log.debug('v6 {}'.format(url))
+                response = self.clnt.get(url, timeout=self.request_timeout)
+                return response
+        return self.clnt.get(url, timeout=self.request_timeout)
+
+    def change_control_approve(self, cc_id, notes="", approve=True):
+        ''' Approve/Unapprove a change control using Resource APIs.
+
```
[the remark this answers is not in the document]
noredistribution: you're right, I was using the wrong `time` key, will be fixed in the next commit
tested it now to approve/unapprove multiple times:
```python
>>> clnt.api.change_control_approve(cc_id,approve=True)
{'value': {'key': {'id': 'bc252f36-412c-4dcb-860c-ac1fc7de050a'}, 'approve': {'value': True, 'notes': ''}, 'version': '2021-12-18T17:44:35.475911747Z'}, 'time': '2021-12-18T17:49:47.143942428Z'}
>>> clnt.api.change_control_approve(cc_id,approve=False)
{'value': {'key': {'id': 'bc252f36-412c-4dcb-860c-ac1fc7de050a'}, 'approve': {'value': False, 'notes': ''}, 'version': '2021-12-18T17:44:35.475911747Z'}, 'time': '2021-12-18T17:49:49.769073197Z'}
>>> clnt.api.change_control_approve(cc_id,approve=True)
{'value': {'key': {'id': 'bc252f36-412c-4dcb-860c-ac1fc7de050a'}, 'approve': {'value': True, 'notes': ''}, 'version': '2021-12-18T17:44:35.475911747Z'}, 'time': '2021-12-18T17:49:57.178556167Z'}
```

cvprac/cvp_api.py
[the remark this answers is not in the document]
noredistribution: oh I think I know what the problem was, should've added an else statement for the last return, as that should be only run if `self.clnt.is_cvaas` is true so I changed it to 
```
if not self.clnt.is_cvaas:
            if self.clnt.apiversion is None:
                self.get_cvp_info()
            if self.clnt.apiversion >= 6.0:
                self.log.debug('v6 {}'.format(url))
                response = self.clnt.get(url, timeout=self.request_timeout)
                return response
        else:
            return self.clnt.get(url, timeout=self.request_timeout)
```
now it returns `None`

```
>>> clnt.apiversion
5.0
>>> tasks = ["2010"]
>>> clnt.api.create_change_control_for_tasks(tasks
KeyboardInterrupt
>>> cc_id = str(uuid.uuid4())
>>> name = f"Change_{datetime.now().strftime('%Y%m%d_%H%M%S')}"
>>> print(clnt.api.change_control_create_for_tasks(cc_id, name, tasks, series=False))
None
```
I guess it'd be better to return `API call not supported on current CVP Version` or something like that?

cvprac/cvp_api.py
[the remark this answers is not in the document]
noredistribution: added error handling for this now

cvprac/cvp_api.py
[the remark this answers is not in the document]
noredistribution: added warning level logging for all resource APIs

cvprac/cvp_api.py
```diff
@@ -3206,4 +3242,470 @@ def workspace_build_status(self, workspace_id, build_id):
...
+            Args:
+              cc_id (str): The ID of the change control.
+              notes (str): An optional approval note.
+              approve (bool): Set to True to approve a change and to False to unapprove a change. The default is True.
+        '''
+        url = '/api/resources/changecontrol/v1/ApproveConfig'
+        # For on-prem check the version as it is only supported from 2021.2.0+
+        # Since the get_change_control already checks this, no need to check it again
+        version = self.change_control_get_one(cc_id)['value']['change']['time']
+        payload = {
```
guillaumeVilar, line 3389: I think it's still necessary to check the version as this function will just return `None`. Thus, when trying this function on 2021.1.1 version I got a `object not subscrictable` error message: 
```
clnt_hume.api.change_control_approve(cc_id='WzkdFQlBz', notes='salut', approve=True)
```
Result: 
```
TypeError                                 Traceback (most recent call last)
/var/folders/hd/3g0v07bx6vd9n3tn3ztkc5900000gn/T/ipykernel_20210/2122868010.py in <module>
----> 1 clnt_hume.api.change_control_approve(cc_id='WzkdFQlBz', notes='salut', approve=True)

~/Documents/tmp/cvprac/cvprac/cvp_api.py in change_control_approve(self, cc_id, notes, approve)
   3386         # For on-prem check the version as it is only supported from 2021.2.0+
   3387         # Since the get_change_control already checks this, no need to check it again
-> 3388         version = self.change_control_get_one(cc_id)['value']['change']['time']
   3389         payload = {
   3390             "key": {

TypeError: 'NoneType' object is not subscriptable
```

cvprac/cvp_api.py
```diff
@@ -3206,4 +3242,470 @@ def workspace_build_status(self, workspace_id, build_id):
...
+                                     ' from 2021.2.0 or newer.')
+                    return None
+        else:
+            return self.clnt.post(url, data=payload, timeout=self.request_timeout)
+
+    def change_control_start(self, cc_id, notes=""):
+        ''' Start a Change Control using Resource APIs.
+            Supported versions: CVP 2021.2.0 or newer and CVaaS.
+            Args:
+                cc_id (string): The ID for the new change control.
```
guillaumeVilar, line 3602: Trying to start a non-existing CC gives the error message: `{"code":9, "message":"not approved"}`
```
print(clnt_imp.api.change_control_start('non-existing-cc-id', notes="my_note"))
```
Result: 
```
---------------------------------------------------------------------------
CvpRequestError                           Traceback (most recent call last)
/var/folders/hd/3g0v07bx6vd9n3tn3ztkc5900000gn/T/ipykernel_20210/1392912492.py in <module>
      1 # Non-existing CC
----> 2 print(clnt_imp.api.change_control_start('non-existing-cc-id', notes="my_note"))

~/Documents/tmp/cvprac/cvprac/cvp_api.py in change_control_start(self, cc_id, notes)
   3623             if self.clnt.apiversion >= 6.0:
   3624                 self.log.debug('v6 ' + str(url) + ' ' + str(payload))
-> 3625                 response = self.clnt.post(url, data=payload, timeout=self.request_timeout)
   3626                 return response
   3627             elif self.clnt.apiversion < 6.0:

~/Documents/tmp/cvprac/cvprac/cvp_client.py in post(self, url, data, files, timeout)
    880                     re-instantiate.
    881         '''
--> 882         return self._make_request('POST', url, timeout, data=data, files=files)
    883 
    884     def delete(self, url, data=None, timeout=30):

~/Documents/tmp/cvprac/cvprac/cvp_client.py in _make_request(self, req_type, url, timeout, data, files)
    615                 full_url = self.url_prefix + url
    616             try:
--> 617                 response = self._send_request(req_type, full_url, timeout,
    618                                               data, files)
    619             except CvpApiError as error:

~/Documents/tmp/cvprac/cvprac/cvp_client.py in _send_request(self, req_type, full_url, timeout, data, files)
    767 
    768             try:
--> 769                 self._is_good_response(response, '%s: %s ' %
    770                                        (req_type, full_url))
    771             except CvpSessionLogOutError as error:

~/Documents/tmp/cvprac/cvprac/cvp_client.py in _is_good_response(self, response, prefix)
    415                                                       response.text)
    416                 self.log.error(msg)
--> 417                 raise CvpRequestError(msg)
    418 
    419         if 'LOG OUT MESSAGE' in response.text:

CvpRequestError: POST: https://<IP>:443/api/resources/changecontrol/v1/ChangeControlConfig : Request Error: Bad Request - {"code":9, "message":"not approved"}
```

Not sure if we can improve that, otherwise I think that's not really a big deal to leave it like that. 

I got the same when trying to stop a non-existing CC: 

```
# Stopping the CC
print(clnt_imp.api.change_control_stop('non-existing-cc-id', notes="my_note2"))
```
Result: 
```
CvpRequestError: POST: https://<IP>:443/api/resources/changecontrol/v1/ChangeControlConfig : Request Error: Bad Request - {"code":9, "message":"not approved"}
```

cvprac/cvp_api.py
```diff
@@ -3206,4 +3242,470 @@ def workspace_build_status(self, workspace_id, build_id):
...
+                                     ' from 2021.2.0 or newer.')
+                    return None
+        else:
+            return self.clnt.post(url, data=payload, timeout=self.request_timeout)
+
+
+    def change_control_schedule(self, cc_id, schedule_time, notes=""):
+        ''' Schedule a Change Control using Resource APIs.
+            Supported versions: CVP 2021.2.0 or newer and CVaaS.
+
```
guillaumeVilar, line 3676: I got a strange issue on CVP - I am not able to approve a CC that was scheduled using this function, thus the CC cannot execute at the scheduled time. 

I tried the following functions: 
```
cc_id = str(uuid.uuid4())
name = "Test CC gigi9 - schedule"
tasks = ['581']
schedule_time = '2021-12-27T13:44:00.0Z'
print(cc_id)

print(clnt_imp.api.change_control_create_for_tasks(cc_id, name, tasks, series=False))
print(clnt_imp.api.change_control_schedule(cc_id=cc_id, schedule_time=schedule_time, notes='salut'))
print(clnt_imp.api.change_control_approve(cc_id=cc_id, notes='salut', approve=True))
```
Result: 
```
9b75b09b-3019-4905-9801-93b8f4648412
{'value': {'key': {'id': '9b75b09b-3019-4905-9801-93b8f4648412'}, 'change': {'name': 'Test CC gigi9 - schedule', 'rootStageId': 'root', 'stages': {'values': {'root': {'name': 'root', 'rows': {'values': [{'values': ['stage0']}]}}, 'stage0': {'name': 'stage0', 'action': {'name': 'task', 'timeout': 3000, 'args': {'values': {'TaskID': '581'}}}}}}, 'notes': 'randomString'}}, 'time': '2021-12-27T13:09:56.285818464Z'}
{'value': {'key': {'id': '9b75b09b-3019-4905-9801-93b8f4648412'}, 'schedule': {'value': '2021-12-27T13:44:00Z', 'notes': 'salut'}}, 'time': '2021-12-27T13:09:56.352372036Z'}
{'value': {'key': {'id': '9b75b09b-3019-4905-9801-93b8f4648412'}, 'approve': {'value': True, 'notes': 'salut'}, 'version': '2021-12-27T13:09:56.285818464Z'}, 'time': '2021-12-27T13:09:56.451946721Z'}
```

From the UI, the CC is not approved. 
However, the API call is returning that the CC is approved: 
```
print(clnt_imp.api.change_control_approval_get_one(cc_id=cc_id))
{'value': {'key': {'id': '9b75b09b-3019-4905-9801-93b8f4648412'}, 'approve': {'value': True, 'notes': 'salut'}, 'version': '2021-12-27T13:09:56.285818464Z'}, 'time': '2021-12-27T13:09:56.451946721Z'}
```

* From the UI, I tried to approve manually the CC, but this doesn't seem to work. 
* When the CC is supposed to execute, the CC fails with the error message: 
```
CC completed with error: Error executing stage stage0 : {"errorCode":"604","errorMessage":"Invalid request"}
```


Manually creating, scheduling, and approving a CC works without issue. 

Tested on 2021.2.2 CVP server.

guillaumeVilar (commented):
Thanks, please find below my reviews. 
I tested as well the example in `docs/labs/lab06-provisioning` and they work perfectly. 

We should probably execute the test suite as some of the code in previous function were modified (workspace and tags related functions). I did not test those functions. 

I guess we should add as well some tests for those new change_control functions in a separate PR?

cvprac/cvp_api.py
mharista: I agree with this. The redundancy here is that the function is calling the get_one resource API and it's function also does the version check. Should we make the version an input param to the function instead of calling get_one to try to populate it?

cvprac/cvp_api.py
mharista: This to me seems like a CVP issue? I think CVP should return a more descriptive error. Essentially something like "CC does not exist"

cvprac/cvp_api.py
mharista: This seems like a potential bug with the APIs. Or could this potentially be some sort of user created issue with timestamps/timezones/formats? Also I'm interested in the appropriate order or operations here.

mharista (changes requested):
Quite of few suggestions/comments/questions here. Happy to hop on a call to discuss as I think that would be quick.

cvprac/cvp_api.py
noredistribution: yes this is unfortunately a bug with the API in CVP, I've added a comment in the example that this might only work from 2021.3.0+; I'll up the apiversion for this, I guess it doesnt' make sense to try to schedule a CC if execution will fail

cvprac/cvp_api.py
noredistribution: filed a bug on CV side to correct the error message

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+        except Exception as e:
+            self.log.warning('change_control_get_one: The change control ID does not exist.')
+            return None
+        if 'approve' not in cc_status['value']:
+            self.log.warning("change_control_approval_get_one: The change has not been approved yet."
+                                " A change has to be approved at least once for the 'approve' state to be populated.")
+            return None
+        return self.clnt.get(url, timeout=self.request_timeout)
+
+
```
mharista, line 3315: Remove excess spaces between functions.

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+            if self.clnt.apiversion < 6.0:
+                self.log.warning('change_control_approval_get_one: Change Control Resource APIs are supported'
+                                    ' from 2021.2.0 or newer.')
+                return None
+        try:
+            cc_status = self.change_control_get_one(cc_id)
+        except Exception as e:
+            self.log.warning('change_control_get_one: The change control ID does not exist.')
+            return None
+        if 'approve' not in cc_status['value']:
```
mharista, line 3309: Interesting case here for if we should check for 'value' key as well or risk an exception.

Ex:

if 'value' not in cc_status or 'approve' not in cc_status['value']:
   ....

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+            Args:
+              cc_id (str): The ID of the change control.
+              notes (str): An optional approval note.
+              approve (bool): Set to True to approve a change and to False to unapprove a change. The default is True.
+        '''
+        url = '/api/resources/changecontrol/v1/ApproveConfig'
+        # For on-prem check the version as it is only supported from 2021.2.0+
+        # Since the get_change_control already checks this, no need to check it again
+        try:
+            version = self.change_control_get_one(cc_id)['value']['change']['time']
```
mharista, line 3350: Check response from change_control_get_one to assign version instead of assuming the keys 'value', 'change', and 'time' will be in the response data.

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+        except Exception as e:
```
mharista, line 3351: If we are going to catch an exception here it would be better to raise it or log it than simply dropping it and returning None.

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+            if self.clnt.apiversion >= 6.0:
+                self.log.debug('v6 ' + str(url) + ' ' + str(payload))
+                response = self.clnt.post(url, data=payload)
+                return response
+            elif self.clnt.apiversion < 6.0:
+                    self.log.warning('change_control_create_with_custom_stages: Change Control Resource APIs are supported'
+                                     ' from 2021.2.0 or newer.')
+                    return None
+        else:
+            return self.clnt.post(url, data=payload)
```
mharista, line 3503: Simplify the API version check, response, and return calls for this function to match others.

cvprac/cvp_api.py
```diff
@@ -3204,6 +3210,449 @@ def workspace_build_status(self, workspace_id, build_id):
...
+                response (dict): A dict that contains...
+                Ex: {'value': {'key': {'id':cc_id,
+                      'time': '...'}
+        '''
+        stages = {'values': {'root': {'name': 'root', 'rows': {'values': []}}}}
+        if series == True:
+            for index, task in enumerate(tasks):
+                stage_id = 'stage%d' % index
+                stages['values']['root']['rows']['values'].append({'values':[stage_id]})
+                stages['values'][stage_id] = {'action': {'args': {'values': {'TaskID': task}},'name':'task','timeout':3000},'name':stage_id}
```
mharista, line 3528: Wondering if this timeout of 3000 should be configurable?

mharista (changes requested):
Few more suggestions.

cvprac/cvp_api.py
mharista: Comment from Tamas:

"UI seems to use 3000 always". Will leave this as is for now and make it configurable in future if necessary.

cvprac/cvp_api.py
```diff
@@ -3204,6 +3208,445 @@ def workspace_build_status(self, workspace_id, build_id):
...
+        # try:
+        #     version = self.change_control_get_one(cc_id)['value']['change']['time']
+        # except Exception as e:
+        #     version = None
+        #     return None
+        cc_status = self.change_control_get_one(cc_id)
+        if cc_status is None:
+            return None
+        if 'value' in cc_status and 'change' in cc_status['value'] and 'time' in cc_status['value']['change']:
+            version = cc_status['value']['change']['time']
```
mharista, line 3355: This is one of the cases where I'm wondering if we want to return None and/or log a message if the version isn't found in the cc_status data?

cvprac/cvp_api.py
mharista: I think we shoould check for value in cc_status as a mentioned.

mharista (changes requested):

mharista (changes requested):
Everything looks good just the one comment about removing extra blank lines between functions.

mharista (approved):
Looks good

Issue #190, opened 2022-03-28
get_configlets should not run get_configlet_by_name for each configlet


noredistribution:
Seems like there was a bug at some point on the CV side due to which the content of the configlets were not getting added to the result when calling getConfiglets.do so to work around it we've used `getConfigletByName.do` and added the `config` to each configlet in the dictionary:

```
        if self.clnt.apiversion == 1.0:
            self.log.debug('v1 Inventory API Call')
            return configlets
        else:
            self.log.debug('v2 Inventory API Call')
            # New API getConfiglets does not return the actual configlet config
            # Get the actual configlet config using getConfigletByName
            if 'data' in configlets:
                for configlet in configlets['data']:
                    full_cfglt_data = self.get_configlet_by_name(
                        configlet['name'])
                    configlet['config'] = full_cfglt_data['config']
            return configlets
```
I've checked 2018.2.5, 2019.1.4, 2020.1.1, 2021.1.1, 2021.2.2, 2021.3.0 and CVaaS and getConfiglets.do works fine and has the config key with the full config, so we should remove this condition or add another condition that it works fine with v3+ aka 2019.1.0+ (to me it worked even with v2, at least with the latest version which is 2018.2.5)

this operation makes the call very slow at scale and instead of a few seconds it can take several minutes to get the final result back

mharista:
Fixed #192

PR #192, opened 2022-04-04
apis: don't run getConfigletByName for every configlet in get_configl…

noredistribution:
…ets function

We've noticed that the time it takes to run get_configlets() using go-cvprac vs python cvprac is 40-50x faster, which seems to have been cause by the fact that in python cvprac we were running getConfigletByName for each configlet to populate the configlets content. I've tested this in quite many CVP versions and do not see this issue where the `conifg` key is empty as I mentioned in #190 I made a small changed and ignored doing the `get_configlet_by_name` for v1 and v4+ apiversions (in theory it should be even with v2, but we don't support those older CV version anymore). On my highest scale system I have about 2000 configlets and here are the differences in exec time:

Before

```
time python3 configlet_test.py
python3 configlet_test.py  6.50s user 1.88s system 1% cpu 13:03.20 total
```

After

```
time python3 configlet_test.py
python3 configlet_test.py  0.89s user 0.89s system 9% cpu 18.720 total
```

so that's 783 seconds (13 min 3 sec) vs 18 seconds, quite a big improvement!

used https://github.com/aristanetworks/cvprac/blob/develop/docs/labs/lab03-configlet-management/backup_configlets.py to test the timing

the only configs that would be empty are builders, which are not actually considered configlets as they are python scripts and if someone wants to retrieve the builder contents there are other specific APIs for that

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -367,8 +367,8 @@ class CvpApi(object):
         configlets = self.clnt.get('/configlet/getConfiglets.do?'
                                    'startIndex=%d&endIndex=%d' % (start, end),
                                    timeout=self.request_timeout)
-        if self.clnt.apiversion == 1.0:
-            self.log.debug('v1 Inventory API Call')
+        if self.clnt.apiversion == 1.0 or self.clnt.apiversion >= 4.0:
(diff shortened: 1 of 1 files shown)


mharista (approved):

mharista:
Fixes #190

PR #193, opened 2022-04-25
test: Adding test for device_decommissioning

noredistribution:
Added test for decommissioning devices post CVP 2021.3.0 to the existing `test_api_inventory` function (note that it takes around 2-3 minutes for decomissioning to be successfuly)

when you run the test with `make systest` you should see: `Verify add_device_to_inventory and delete_device(s) ... ok`

### DIFF
--- a/test/system/test_cvp_client_api.py
+++ b/test/system/test_cvp_client_api.py
@@ -1669,9 +1669,18 @@ class TestCvpClient(TestCvpClientBase):
         # Get devices current configlets
         orig_configlets = self.api.get_configlets_by_device_id(device['key'])
         # delete from inventory
-        self.api.delete_device(device['systemMacAddress'])
-        # sleep to allow delete to complete
-        time.sleep(1)
+        if self.clnt.apiversion <= 7.0:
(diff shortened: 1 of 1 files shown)


test/system/test_cvp_client_api.py
```diff
@@ -1669,9 +1669,18 @@ def test_api_inventory(self):
...
+            # sleep to allow delete to complete
+            time.sleep(1)
+        else:
+            request_id = str(uuid.uuid4())
+            self.api.device_decommissioning(device['serialNumber'], request_id)
+            decomm_status = "DECOMMISSIONING_STATUS_SUCCESS"
+            decomm_result = ""
+            while decomm_result != decomm_status:
+                decomm_result = self.api.device_decommissioning_status_get_one(request_id)['value']['status']          
+                time.sleep(10)
```
mharista, line 1683: Maybe we should have a timeout for cases where an unexpected issue arises? Right now this will loop forever unless the decommission API call has a failure. Maybe 10 minutes as a timeout?

mharista (changes requested):

mharista (approved):

PR #201, opened 2022-05-24
Raise JSONDecodeError instead of returning a string.

### LINKED ISSUE #200: cvprac may return an unterminated string
When trying to parse the received payload from CVP, requests may fail to parse it as JSON. As a cvprac user, I would expect that an exception is raised. The current implementation catches `requests.exceptions.JSONDecodeError` and returns an unterminated string.
Logs when this is happening: `requests.exceptions.JSONDecodeError: [Errno Unterminated string starting at] {"data": [TRUNCATED]`

Error handling logic [here](https://github.com/aristanetworks/cvprac/blob/develop/cvprac/cvp_client.py#L671) should be modified.

mtache:
Fixes #200

### DIFF
--- a/cvprac/cvp_client.py
+++ b/cvprac/cvp_client.py
@@ -100,7 +100,7 @@ from pkg_resources import parse_version
 
 import requests
 from requests.exceptions import ConnectionError, HTTPError, Timeout, \
-    ReadTimeout, TooManyRedirects
+    ReadTimeout, TooManyRedirects, JSONDecodeError
 
 from cvprac.cvp_api import CvpApi
 from cvprac.cvp_client_errors import CvpApiError, CvpLoginError, \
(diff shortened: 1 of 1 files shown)


cvprac/cvp_client.py
```diff
@@ -660,17 +660,18 @@ def _make_request(self, req_type, url, timeout, data=None,
...
                 if 'Extra data' in str(error):
                     self.log.debug('Found multiple objects in response data.'
                                    'Attempt to decode')
                     decoded_data = json_decoder(response.text)
                     resp_data = dict(data=decoded_data)
                 else:
-                    self.log.debug('Attempt to return response text')
-                    resp_data = dict(data=response.text)
+                    self.log.error(error)
+                    raise error
```
mharista, line 674: This breaks handling of Resource API endpoints that return multiple JSON objects.

The specific format that isn't being handled is below and it was likely previously masked by the response.text.

Bad Format not being handled:

[Errno Expecting value] : 0

Good format:

[Errno Extra data] {"result":{"value":{"key":{"workspaceId":"CVPRAC_TEST_b7e898b8-dba8-11ec-ba4c-a683e7158098","elementType":"ELEMENT_TYPE_DEVICE","label":"cvpractestdev","value":"TAGTESTDEV"},"remove":false},"time":"2022-05-24T21:30:23.853711633Z","type":"INITIAL"}}
{"result":{"value":{"key":{"workspaceId":"CVPRAC_TEST_b7e898b8-dba8-11ec-ba4c-a683e7158098","elementType":"ELEMENT_TYPE_INTERFACE","label":"cvpractestint","value":"TAGTESTINT"},"remove":false},"time":"2022-05-24T21:30:23.710580226Z","type":"INITIAL"}}
: 249

mharista (changes requested):
Need to dig into this odd format returned by Resource API endpoints.

mharista:
Resolved via #202

PR #212, opened 2022-09-13
Example: Add example on how to automatically reconcile configurations

noredistribution:
in this example you can see how to reconcile the configuration of all out of configuration compliance devices

### DIFF
--- a/docs/labs/lab06-provisioning/auto_reconcile_on_rc_change.py
+++ b/docs/labs/lab06-provisioning/auto_reconcile_on_rc_change.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2022 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the COPYING file.
+
+# This script can be run as a cronjob to periodically reconcile all devices
+# that are out of configuration compliance in environments where the running-config
+# is still modified via the CLI often.
+from cvprac.cvp_client import CvpClient
+import ssl
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


docs/labs/lab06-provisioning/auto_reconcile_on_rc_change.py
```diff
@@ -0,0 +1,64 @@
...
+        print(device['hostname'])
+        dev_mac = device['systemMacAddress']
+        # check if device already has reconciled config and save the key if it does
+        try:
+            configlets = clnt.api.get_configlets_by_device_id(dev_mac)
+            for configlet in configlets:
+                if configlet['reconciled'] == True:
+                    configlet_key = configlet['key']
+                else:
+                    configlet_key = ""
```
mharista, line 53: This appears to overwrite/clear the reconcile configlet 'configlet_key' if the reconcile configlet is not the last configlet in the full list of configlets.

mharista (changes requested):

docs/labs/lab06-provisioning/auto_reconcile_on_rc_change.py
noredistribution: in our case the ask was to have one reconcile configlet per device and keep updating that after the customer changes the running-config manually on the CLI
if there's no reconciled configlet yet on the device then I have to create a new configlet, but if there is one already then I just need to update the existing one so have to fetch the configlet key for that

mharista (approved):

PR #216, opened 2022-09-21
Update get_running_configs_by_time.py

noredistribution:
in some CVP versions the config will not be in the first index, looping through the list and setting the result only if 'config' key exists in any of the indexes

### DIFF
--- a/docs/labs/lab02-inventory-operations/get_running_configs_by_time.py
+++ b/docs/labs/lab02-inventory-operations/get_running_configs_by_time.py
@@ -23,8 +23,11 @@ for device in inventory:
         }
     }
     try:
-        resultRunningConfig = clnt.post(uri, data=data)[0]['config']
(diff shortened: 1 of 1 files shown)


docs/labs/lab02-inventory-operations/get_running_configs_by_time.py
```diff
@@ -23,8 +23,11 @@
         }
     }
     try:
-        resultRunningConfig = clnt.post(uri, data=data)[0]['config']
+        resultRunningConfig = clnt.post(uri, data=data)
+        for idx in resultRunningConfig:
+            if 'config' in idx:
+                result = idx['config']
```
mharista, line 29: In this scenario would it make sense to break the loop once the 'config' key is found and the result is set?

mharista (changes requested):

mharista (approved):

PR #229, opened 2022-10-14
Examples: Adding an e2e provisioning example for ATD and updating the…

noredistribution:
… tag assignment example

### DIFF
--- a/docs/labs/lab08-resource-apis/topology_tag_assignment.py
+++ b/docs/labs/lab08-resource-apis/topology_tag_assignment.py
@@ -80,7 +80,7 @@ for tag in tags_common+tags_spines:
 request = 'REQUEST_START_BUILD'
 request_id = 'b1'
 description='testing cvprac build'
-build_ws = clnt.api.workspace_config(workspace_id=workspace_id, display_name=display_name, 
+clnt.api.workspace_config(workspace_id=workspace_id, display_name=display_name, 
                                      description=description, request=request, request_id=request_id)
 
 ### Check workspace build status and proceed only after it finishes building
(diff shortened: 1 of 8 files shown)


docs/labs/lab08-resource-apis/topology_tag_assignment.py
```diff
@@ -2,7 +2,22 @@
 # Use of this source code is governed by the Apache License 2.0
 # that can be found in the COPYING file.
 
+# In this example we are going to assign topology tags using the tags.v2 workspace aware API
+# More details on tag.v2 can be found at https://aristanetworks.github.io/cloudvision-apis/models/tag.v2/
+# NOTE: Tag.v2 can be used for assigning both device and interface tags (studios, topology, etc) and it's not
+# limited to topology tags only. 
+# The following are built-in tags that can be used to modify the Topology rendering:
+# topology_hint_type: < core | edge | endpoint | leaf | management | spine | leaf >
```
mharista, line 10: "leaf" is listed twice in the options list.

mharista (changes requested):
One nit pick comment. After that, as long as Dharmesh's testing is good, I think this is good to go.

docs/labs/lab08-resource-apis/topology_tag_assignment.py
```diff
@@ -62,13 +80,21 @@
 request = 'REQUEST_START_BUILD'
 request_id = 'b1'
 description='testing cvprac build'
-clnt.api.workspace_config(workspace_id=workspace_id,display_name=display_name,description=description,request=request,request_id=request_id)
+build_ws = clnt.api.workspace_config(workspace_id=workspace_id, display_name=display_name,
```
mharista, line 83: Why add assignment of this build_ws variable if it is never used?

mharista (changes requested):

docs/labs/lab08-resource-apis/topology_tag_assignment.py
noredistribution: hmm you're right, let me change that back

mharista (approved):

PR #235, opened 2022-12-14
Fix: Use status endpoint instead of config endpoint for service account resource APIs

### LINKED ISSUE #234: Feat: Add serviceaccount.v1.AccountService rAPI
Currently the svc_account_get_all() only fetches the config state of the service accounts and there's a use case to fetch the status instead which contains the `created_by` field as well. That way we can do a GetAll and figure out which svc account was created by whom so if a user is deleted we could also delete the service accounts that the user created.

noredistribution:
Fixes #234 

Change `svc_account_get_one` and `svc_account_get_all` to use the status endpoints instead of the config endpoints

`AccountService/GetOne` and `AccountService/GetAll` instead of `AccountConfigeService/GetOne` and `AccountConfigService/GetAll`

this will gain us two extra key values:
- `created_by`: the user that created the service account
- `last_access`: the last time the service account accessed the system

this change should be transparent for users who've used these functions before as the return data has the same format + those 2 k,v pairs

- added system test as well to check for those 2 new keys

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -3875,7 +3875,7 @@ class CvpApi(object):
             return self.clnt.post(url, data=payload)
 
     def svc_account_get_all(self):
-        ''' Get all service account config/status states using Resource APIs.
+        ''' Get all service account states using Resource APIs.
             Supported versions: CVP 2021.3.0 or newer and CVaaS.
             Returns:
                 response (list): Returns a list of dictionaries that contains...
@@ -3891,7 +3891,7 @@ class CvpApi(object):
             return self.clnt.post(url)
 
     def svc_account_get_one(self, username):
-        ''' Get a service account's config/status state using Resource APIs
+        ''' Get a service account's state using Resource APIs
             Supported versions: CVP 2021.3.0 or newer and CVaaS.
             Args:
                 username (string): The service account username.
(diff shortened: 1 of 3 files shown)


docs/labs/lab07-aaa/delete_svc_account_created_by_user.py
```diff
@@ -0,0 +1,23 @@
...
+clnt = CvpClient()
+clnt.connect(['cvp1'],'username', 'password')
+
+svc_accounts = clnt.api.svc_account_get_all()
+created_by = 'john.smith'
+
+# Delete service accounts created by user john.smith
+for account in svc_accounts:
+    if account['value']['created_by'] == created_by:
+        acc_to_delete = account
```
mharista, line 22: the `acc_to_delete` variable isn't needed.

cvprac/cvp_api.py
```diff
@@ -3875,7 +3875,7 @@ def svc_account_token_set(self, username, duration, description):
             return self.clnt.post(url, data=payload)
 
     def svc_account_get_all(self):
-        ''' Get all service account states using Resource APIs.
+        ''' Get all service account config/status states using Resource APIs.
```
mharista, line 3878: Since we are switching from the service account Config Resource API to the service account State Resource API can we update this doc string and the get_one below to just say "Get all service accounts state using Resource APIs" and "Get a service account's state using Resource APIs"

mharista (changes requested):

mharista (approved):

PR #238, opened 2023-01-18
Feat: Add functions for add_role, get_role, update_role and delete_role API endpoints

vmmor:
Check if comments are OK or if we should add examples as well.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -3790,15 +3790,71 @@ class CvpApi(object):
             self.log.debug('v7 ' + str(url))
             return self.clnt.post(url, data=payload, timeout=self.request_timeout)
 
+    def add_role(self, name, description, moduleList):
+        ''' Add new local role to the CVP UI.
+            Args:
+                name (str): local role name on CVP
+                description (str): role description
+                moduleList (list): list of modules (name (str) and mode (str))
+        '''
+        data = {"name": name,
+                "description": description,
+                "moduleList": moduleList}
+        return self.clnt.post('/role/createRole.do', data=data,
+                              timeout=self.request_timeout)
+
+    def update_role(self, rolekey, name, description, moduleList):
+        ''' Updates role information, like
+            role name, description and role modules.
+            Args:
+                rolekey (str): local role key on CVP
+                name (str): local role name on CVP
+                description (str): role description
+                moduleList (list): list of modules (name (str) and mode (str))
+        '''
+        data = {"key": rolekey,
+                "name": name,
+                "description": description,
+                "moduleList": moduleList}
+        return self.clnt.post('/role/updateRole.do', data=data,
+                              timeout=self.request_timeout)
+
+    def get_role(self, rolekey):
+        ''' Returns specified role information.
+            Args:
+                rolekey (str): role key on CVP
+            Returns:
+               response (dict): Returns a dict that contains the role.
+               Ex: {'name': 'Test Role', 'key': 'role_1599019487020581247', 'description': 'Test'...}
+        '''
+        return self.clnt.get('/role/getRole.do?roleId={}'.format(rolekey),
+                             timeout=self.request_timeout)
+
     def get_roles(self):
         ''' Get all the user roles in CloudVision.
             Returns:
                response (dict): Returns a dict that contains all the user role states..
                Ex: {'total': 7, 'roles': [{'name': 'Test Role', 'key': 'role_1599019487020581247',
-               'description': 'Test'...}
+               'description': 'Test'...}]}
         '''
         url = '/role/getRoles.do?startIndex=0&endIndex=0'
         return self.clnt.get(url, timeout=self.request_timeout)
+
+    def delete_role(self, rolekey):
+        ''' Remove specified role from CVP
+            Args:
+                rolekey (str): role key on CVP
+        '''
+        data = [rolekey]
+        return self.delete_roles(data)
+
+    def delete_roles(self, rolekeys):
+        ''' Remove specified roles from CVP
+            Args:
+                rolekeys (list): list of role keys (str) on CVP
+        '''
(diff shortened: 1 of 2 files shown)


cvprac/cvp_api.py
```diff
@@ -3790,6 +3790,46 @@ def device_decommissioning_status_get_all(self, status="DECOMMISSIONING_STATUS_U
             self.log.debug('v7 ' + str(url))
             return self.clnt.post(url, data=payload, timeout=self.request_timeout)
 
+    def add_role(self, name, description, moduleList):
+        ''' Add new local role to the CVP UI.
+            Args:
+                name (str): local role name on CVP
+                description (str): role description
+                moduleList (str): list of modules (name (str) and mode (str))
```
mharista, line 3798: Is this a list object or a string with some specific list formatting?

cvprac/cvp_api.py
```diff
@@ -3790,6 +3790,46 @@ def device_decommissioning_status_get_all(self, status="DECOMMISSIONING_STATUS_U
...
+                              timeout=self.request_timeout)
+
+    def update_role(self, rolekey, name, description, moduleList):
+        ''' Updates role information, like
+            role name, description and role modules.
+            Args:
+                rolekey (str): local role key on CVP
+                name (str): local role name on CVP
+                description (str): role description
+                moduleList (str): list of modules (name (str) and mode (str))
```
mharista, line 3813: Same as above.

cvprac/cvp_api.py
```diff
@@ -3800,6 +3840,15 @@ def get_roles(self):
...
 
+    def delete_role(self, rolekey):
+        ''' Remove specified role from CVP
+            Args:
+                rolekey (str): role key on CVP
+        '''
+        data = [rolekey]
+        return self.clnt.post('/role/deleteRoles.do', data=data,
+                              timeout=self.request_timeout)
+
```
mharista, line 3851: Let's create a function for delete_roles() that takes a list of role keys and then another function for delete_role that will take one role key as string and call delete_roles() where the key is placed into a list.

test/system/test_cvp_client_api.py
```diff
@@ -221,6 +221,119 @@ def test_api_user_operations(self):
...
+        # pylint: disable=too-many-branches
+        # Test Get All Roles
+        result = self.api.get_roles()
+        self.assertIsNotNone(result)
+        self.assertIn('total', result)
+        start_total = result['total']
+
+        # Test Create Role
+        result = self.api.add_role('test_cvp_role', 'role description',
+                                    [{'name': 'image', 'mode': 'rw'}])
```
mharista, line 237: I think this answers my question above about the type for moduleList. Likely the fix is to update the type for moduleList in the doc strings above from str to list.

test/system/test_cvp_client_api.py
```diff
@@ -221,6 +221,119 @@ def test_api_user_operations(self):
...
+        self.assertIsNotNone(result)
+        self.assertIn('data', result)
+        self.assertIn('key', result['data'])
+        self.assertIsNotNone(result['data']['key'])
+        self.assertIn('name', result['data'])
+        self.assertEqual(result['data']['name'], 'test_cvp_role')
+        self.assertIn('description', result['data'])
+        self.assertEqual(result['data']['description'], 'role description')
+        self.assertIn('createdBy', result['data'])
+        self.assertEqual(result['data']['createdBy'], 'cvpadmin')
```
mharista, line 247: Likely don't want to hardcode the username check here since it is configurable in cvp_nodes.yaml.

test/system/test_cvp_client_api.py
```diff
@@ -221,6 +221,119 @@ def test_api_user_operations(self):
...
+        result = self.api.get_role(new_role_key)
+        self.assertIsNotNone(result)
+        self.assertIn('key', result)
+        self.assertEqual(result['key'], new_role_key)
+        self.assertIn('name', result)
+        self.assertEqual(result['name'], 'test_cvp_role')
+        self.assertIn('description', result)
+        self.assertEqual(result['description'], 'role description')
+        self.assertIn('createdBy', result)
+        self.assertEqual(result['createdBy'], 'cvpadmin')
```
mharista, line 266: Same as above.

test/system/test_cvp_client_api.py
```diff
@@ -221,6 +221,119 @@ def test_api_user_operations(self):
...
+        result = self.api.get_role(new_role_key)
+        self.assertIsNotNone(result)
+        self.assertIn('key', result)
+        self.assertEqual(result['key'], new_role_key)
+        self.assertIn('name', result)
+        self.assertEqual(result['name'], update_role_name)
+        self.assertIn('description', result)
+        self.assertEqual(result['description'], update_role_description)
+        self.assertIn('createdBy', result)
+        self.assertEqual(result['createdBy'], 'cvpadmin')
```
mharista, line 307: Same as above.

mharista (changes requested):

cvprac/cvp_api.py
vmmor: It's a list. Updated.

cvprac/cvp_api.py
vmmor: It's a list. Updated.

cvprac/cvp_api.py
vmmor: done. added tests for both delete_role and delete_roles.

test/system/test_cvp_client_api.py
vmmor: changed the test to just check that the createdBy is not none.

test/system/test_cvp_client_api.py
vmmor: changed the test to just check that the createdBy is not none.

test/system/test_cvp_client_api.py
vmmor: changed the test to just check that the createdBy is not none.

PR #248, opened 2023-03-24
Fix: Add workaround for CVP API validation config

chetryan:
Adds a workaround for the case where CVP API splits warnings over several strings when validating
configuration that has both errors and warnings.

# Changes
- add a sanitization method to cvp_api & pass validation reply through sanitization
- add unittests
- add sytemtests

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -38,6 +38,7 @@ import time
 # pylint: disable=redefined-builtin
 from io import open
 from datetime import datetime
+from re import split
 
 from cvprac.cvp_client_errors import CvpApiError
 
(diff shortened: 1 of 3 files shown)


cvprac/cvp_api.py
```diff
@@ -1308,6 +1310,51 @@ def add_note_to_configlet(self, key, note):
...
+        # We capture as well the remaining \\n or whitespace and include
+        # the extra ', ' added in the previous step in the matching criteria.
+        # The extra ', ' is not included in the strings of the new list
+        temp_warnings = split(
+            r'(.*?at line \d+.*?),\s+',
+            temp_warnings
+        )
+
+        # The behaviour of re.split will add empty strings
+        # if the regex matches on the begging or ending of the lin.
```
mharista, line 1348: Typo in comment? line?

mharista (changes requested):

cvprac/cvp_api.py
chetryan: Correct, this is a typo. Update this line in the latest commit.

mharista (approved):

PR #252, opened 2023-04-21
Test: Fix test formats for cvprac pipeline environment and until better version handling is implemented

mharista:


test/system/test_cvp_client_api.py
```diff
@@ -808,53 +808,34 @@ def test_api_get_configlets_and_mappers(self):
...
-                                    if 'Entity does not exist' in e.msg:
-                                        # Configlet Builder for 2022.3.X+
-                                        cfglt = self.api.get_configlet_by_name(
-                                            'SYS_TelemetryBuilderV6')
-                    else:
-                        raise
-            else:
-                raise
-        result = self.api.get_configlet_builder(cfglt['key'])
+        cfglt_bldr_key = None
```
chetryan, line 811: After testing with the develop version, I don't think we need these changes any more.

test/system/test_cvp_client_api.py
```diff
@@ -808,53 +808,34 @@ def test_api_get_configlets_and_mappers(self):
...
-        # returned by the request
-        exp_data = {
-            'isAssigned': bool,
-            'formList': list,
-            'main_script': dict,
-        }
-        # Handle unicode type for Python 2 vs Python 3
-        if sys.version_info.major < 3:
-            exp_data[u'name'] = (unicode, str)
+            # Verify the following keys and types are
```
chetryan, line 821: I don't think is required anymore

chetryan (commented):
I don't think all changes are required.

test/system/test_cvp_client_api.py
mharista: Keeping this update for simplification and better format for going foward.

test/system/test_cvp_client_api.py
mharista: Will remove and test. A nice first step for Python2 deprecation.

test/system/test_cvp_client_api.py
chetryan: Discussed that it is best if that this change remainss.

test/system/test_cvp_client_api.py
chetryan: Discussed that it is best if that this change remains.

PR #253, opened 2023-07-13
Fix: ability to use device_decommissioning for unprovisioned devices

noredistribution:
## Description
In the current version of cvprac the `device_decommissioning()` function uses `get_device_by_serial()` to check if the device exists in CloudVision however that API works strictly for provisioned devices, so if you have a device that is not provisioned you cannot decommission it and you'd get the following error

```python
>>> clnt.api.device_decommissioning(device_id, request_id)
2023-Jul-13 21:33:06 IST <WARNING> cvp_api device_decommissioning: Device with JPE19290408 serial number does not exist (or is not registered) to decommission
```

to improve on this instead of using `get_device_by_serial()`, I've used `get_inventory()`, however in its current state `get_inventory()` had the `provisioned` query parameter always set to True, so I've added a positional argument at the end of `get_inventory()` set `provisioned=True` by default, and then in the `device_decommissioning()` I'm calling the `get_inventory(provisioned=False)` to fetch all devices in the inventory of CloudVision. This will enable decommissioning every device regardless of their streaming status and provisioning status.

## Test

Haven't added additional tests as this should be transparent for any current script.
To locally test the following lab example can be used: 
https://github.com/aristanetworks/cvprac/blob/develop/docs/labs/lab02-inventory-operations/remove_and_decommission_device.py

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -572,7 +572,7 @@ class CvpApi(object):
                              '%s&queryparam=&startIndex=%d&endIndex=%d' %
                              (key, start, end), timeout=self.request_timeout)
 
-    def get_inventory(self, start=0, end=0, query=''):
+    def get_inventory(self, start=0, end=0, query='', provisioned=True):
         ''' Returns the a dict of the net elements known to CVP.
 
             Args:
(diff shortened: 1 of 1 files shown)


mharista (approved):
Looks good to me. I have a systest env setup now and will merge and validate.

PR #255, opened 2023-08-26
Feat: add support for config validation during config assign

noredistribution:
Currently when using `apply_configlets_to_device()` when `create_task=False` we actually just ommit saving topology, like a pre-staging operation, however if there are no diffs between designed-config and running-config, we should just do what the UI does and not generate a task at all by running validateAndCompareConfiglets.do and add `configCompareCount` dict into the addTempAction.do payload. If all keys in that dict, namely `new`, `mismatch` and `reconcile` are 0, the backend will not generate a task.

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -1463,7 +1463,7 @@ class CvpApi(object):
         return self.clnt.post(url, data=data, timeout=self.request_timeout)
 
     def apply_configlets_to_device(self, app_name, dev, new_configlets,
-                                   create_task=True, reorder_configlets=False):
+                                   create_task=True, reorder_configlets=False, validate=False):
         ''' Apply the configlets to the device.
 
             Args:
@@ -1484,6 +1484,12 @@ class CvpApi(object):
                     directly. Set this parameter to True only with the full
                     list of configlets being applied to the device provided
                     via the new_configlets parameter.
+                validate (bool): Defaults to False. If set to True, the function
+                    will validate and compare the configlets to be attached and
+                    pupulate the configCompareCount field in the data dict. In case
+                    all keys are 0, ie there is no difference between designed-config
+                    and running-config after applying the configlets, no task will be
+                    generated.
 
             Returns:
                 response (dict): A dict that contains a status and a list of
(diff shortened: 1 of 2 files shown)


cvprac/cvp_api.py
```diff
@@ -1484,6 +1484,12 @@ def apply_configlets_to_device(self, app_name, dev, new_configlets,
                     directly. Set this parameter to True only with the full
                     list of configlets being applied to the device provided
                     via the new_configlets parameter.
+                validate (bool): Defaults to False. If set to True, the function
+                    will validate and compare the configlets to be attached and
+                    pupulate the configCompareCount field in the data dict. In case
```
mharista, line 1489: typo: populate.

cvprac/cvp_api.py
```diff
@@ -1536,6 +1542,19 @@ def apply_configlets_to_device(self, app_name, dev, new_configlets,
...
+        if validate:
+            validation_result = self.validate_configlets_for_device(dev['systemMacAddress'], ckeys)
+            mismatch_count = validation_result['mismatch']
+            new_count = validation_result['new']
+            reconcile_count = validation_result['reconcile']
+            data['data'][0].update({
+                "configCompareCount": {
+                    "mismatch": mismatch_count,
+                    "reconcile": reconcile_count,
+                    "new": new_count
```
mharista, line 1554: This can be simplified to remove the unnecessary variables being instantiated.

test/system/test_cvp_client_api.py
```diff
@@ -1532,6 +1533,29 @@ def test_api_configlets_to_device(self):
...
+            label,
+            self.device,
+            [new_configlet_data],
+            validate=True
+        )
+
+        # Check that no taskids have been generated
+        self.assertEqual(validate_and_apply['data']['taskIds'], [])
+        self.assertEqual(validate_and_apply['data']['status'], 'success')
+
```
mharista, line 1558: we should remove and delete the new _rc configlet after the test runs.

mharista (changes requested):
Few changes. Overall this is a great feature to add. Thanks!

mharista (approved):
Good to go

PR #256, opened 2023-08-30
Feat: add support for config validation during config removal

noredistribution:
This is the continuation of PR#255 added the same validation on config removal, which avoids generating empty task for no-op changes


mharista (approved):
Looks good!

PR #258, opened 2023-10-12
FIX: Add check to connect() to ensure token works

chetryan:
# Summary
At the moment, cvprac does not do any validation during `connect()` if `api_token` is given.

# Tickets
- Currently impacts the `ansible-cvp` collection : https://github.com/aristanetworks/ansible-cvp/issues/675

# Impact. 
`connect()` does not raise any error for invalid, expired or incorrect token. Thus upstream packages are assuming tokens are valid, leading to errors down the line.

# Workaround
None

# Fix
- Added a request to the cvpInfo that should only return 200 if the token is valid and working.
- Added system tests

### DIFF
--- a/test/fixtures/cvp_nodes.yaml
+++ b/test/fixtures/cvp_nodes.yaml
@@ -10,3 +10,7 @@
   username: CvpRacTest
   password: AristaInnovates
   device: python-test-2
+  # The below fields can be definded to test the token test cases.
+  # If they are undefined, the tests are skipped
+  # api_token: <token here>
+  # api_token_expired: <expired token here>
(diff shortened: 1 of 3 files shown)


test/fixtures/cvp_nodes.yaml
```diff
@@ -10,3 +10,7 @@
   username: CvpRacTest
   password: AristaInnovates
   device: python-test-2
+  # The below fields can be definded to test the token test cases.
```
mharista, line 13: Typo: defined

test/fixtures/cvp_nodes.yaml
```diff
@@ -10,3 +10,7 @@
...
+  # If they are undefined, the tests are skipped
+  # api_token: <token here>
+  # api_token_expired: <expired token here>
```
mharista, line 16: Add new line at end of file

mharista (changes requested):
Two minor nit picks. After that everything looks good to go.

mharista (approved):

PR #261, opened 2023-10-31
Refactor: Move from pkg_resources to packaging

gmuloc:
To support python3.12

```
Python 3.12.0 (main, Oct 25 2023, 09:27:37) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cvprac
>>> from cvprac.cvp_client import CvpClient
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/private/tmp/py312/venv/lib/python3.12/site-packages/cvprac/cvp_client.py", line 99, in <module>
    from pkg_resources import parse_version
ModuleNotFoundError: No module named 'pkg_resources'
>>>
```

https://docs.python.org/3/whatsnew/3.12.html

[gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environments

NOTE: it could have been possible to add setuptools to requirement to continue using pkg_resources but it gives a deprecation warning

```
Python 3.12.0 (main, Oct 25 2023, 09:27:37) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pkg_resources
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html

### DIFF
--- a/cvprac/cvp_client.py
+++ b/cvprac/cvp_client.py
@@ -96,7 +96,7 @@ import json
 import logging
 from logging.handlers import SysLogHandler
 from itertools import cycle
-from pkg_resources import parse_version
+from packaging.version import parse
 
 import requests
 from requests.exceptions import ConnectionError, HTTPError, Timeout, \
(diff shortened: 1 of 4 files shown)


mharista:
Revisiting this as I am working on a new release. Only potential drawback I see here is that `packaging` requires 3.7 minimum. I don't think we have a minimum Python3 version requirement currently. Anyone see any issues with adding this constraint?

mharista (approved):

mharista:
Moved to #271

PR #264, opened 2024-01-30
Fix: missing url encoding for get_user()

noredistribution:
We forgot to add URL encoding for the single getUser.do API, we had it for getUsers.do

Before:

we'd get

```
CvpApiError: GET: https://192.0.2.79:443/web/user/getUser.do?userId=bob@abc.com : Request Error: Entity does not exist
```

instead of 

```
CvpApiError: GET: https://192.0.2.79:443/web/user/getUser.do?userId=bob%40abc.com : Request Error: Entity does not exist
```

or an actual response

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -220,7 +220,7 @@ class CvpApi(object):
             Args:
                 username (str): username on CVP
         '''
-        return self.clnt.get('/user/getUser.do?userId={}'.format(username),
+        return self.clnt.get('/user/getUser.do?userId={}'.format(qplus(username)),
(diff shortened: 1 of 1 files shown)


chetryan:
Is this behaviour only observed on that API call? I see the username being supplied to other calls as well. Should all requests be encoded?

chetryan (approved):
Looks good to me.

chetryan:
> Is this behaviour only observed on that API call? I see the username being supplied to other calls as well. Should all requests be encoded?

Clarified by Tamas that the post to /user/updateUser.do is not impacted and does not need this encoding.

PR #269, opened 2024-03-21
Fix: updating the enrollment endpoint for CVaaS

noredistribution:
On CVaaS we're migrating away from the /api/v3/services endpoints and moving everything to resource APIs, as part of one of the phases we've moved the enrollment endpoint from

`/api/v3/services/admin.Enrollment/AddEnrollmentToken`

to

`/api/resources/admin.Enrollment/AddEnrollmentToken`

the payload stayed the same with the exception that the `validFor` key only accepts seconds now, e.g. 86400s (where previously you could do '24h' or '60m';

 we will also migrate on on-prem once we get to 2024.2.0 (we'll do a separate PR for that in a few months)

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -3061,30 +3061,33 @@ class CvpApi(object):
                devices (list): list of device Serial Numbers for which the
                    token should be generated. The default is all devices.
                duration (string): the token's validity time (max 1 month),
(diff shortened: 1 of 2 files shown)


mharista (approved):

chetryan (approved):
LGTM

PR #271, opened 2024-04-16
Fix: Move from pkg_resources to packaging

mharista:


rewati-arista (approved):
lg, out of curiosity what is the reason to change from pkg_resources.parse_version to packaging.version.parse?

mharista:
@rewati-arista the reasoning is in the old PR #261 

Here is a portion of the info:

https://docs.python.org/3/whatsnew/3.12.html

https://github.com/python/cpython/issues/95299: Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environments

NOTE: it could have been possible to add setuptools to requirement to continue using pkg_resources but it gives a deprecation warning

cvprac/cvp_client.py
dlewis-arista, line 1: Nitpicking so i've gone and approved anyway but could probably store parse(full_version) before the elif block instead of invoking it each stmt, or maybe map the versions to api versions in a data struct and iterate through that to make the comparison.

dlewis-arista (approved):

PR #273, opened 2024-04-24
Py3 lint fix

mharista:
- Fix all Py3 pylint issues
- Fix all Py3 pep8 warnings

- Successfully able to run make all for Py3 now.
- All unit and system tests still passing against CVP 2024.1.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ coverage_report:
 
 pep8:
 	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE) cvprac/
-	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE),E402 test/
+	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE),E402,E501 test/
 
 pyflakes:
 	pyflakes cvprac/ test/
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -1,3 +1,5 @@
+# pylint: disable=fixme
+# pylint: disable=too-many-locals
 #
 # Copyright (c) 2017, Arista Networks, Inc.
 # All rights reserved.
@@ -1429,7 +1431,7 @@ class CvpApi():
         url = '/provisioning/v2/saveTopology.do'
         return self.clnt.post(url, data=data, timeout=self.request_timeout)
 
-    def apply_configlets_to_device(self, app_name, dev, new_configlets,
+    def apply_configlets_to_device(self, app_name, dev, new_configlets, # pylint: disable=too-many-locals
                                    create_task=True, reorder_configlets=False, validate=False):
         ''' Apply the configlets to the device.
 
@@ -1599,7 +1601,8 @@ class CvpApi():
                           'childTasks': [],
                           'parentTask': ''}]}
         if validate:
-            validation_result = self.validate_configlets_for_device(dev['systemMacAddress'], keep_keys)
+            validation_result = self.validate_configlets_for_device(dev['systemMacAddress'],
+                                                                    keep_keys)
             data['data'][0].update({
                 "configCompareCount": {
                     "mismatch": validation_result['mismatch'],
@@ -2059,7 +2062,7 @@ class CvpApi():
         if self.clnt.apiversion is None:
             self.get_cvp_info()
         if self.clnt.apiversion >= 2.0:
-            if resp['complianceIndication'] == u'':
+            if resp['complianceIndication'] == '':
                 resp['complianceIndication'] = 'NONE'
         return resp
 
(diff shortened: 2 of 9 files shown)


cvprac/cvp_api.py
```diff
@@ -3828,11 +3781,9 @@ def device_decommissioning(self, device_id, request_id):
                 url = '/api/resources/inventory/v1/DeviceDecommissioningConfig'
                 self.log.debug('v7 ' + str(url) + ' ' + str(payload))
                 return self.clnt.post(url, data=payload, timeout=self.request_timeout)
-        else:
-            self.log.warning(
-                'Device with %s serial number does not exist (or is not registered) to decommission'
-                % device_id)
-            return None
+        self.log.warning(f"Device with {device_id} serial number does not exist"
```
rewati-arista, line 3784: would we want this warning to be logged even when outer if was met (device exists) but inner wasn't (version compare evaluates to false)?
in earlier logic we would not have but in newer logic we will since we removed the else

not sure if such a flow would happen (lack the context) but putting it out there 👍🏻

cvprac/cvp_api.py
```diff
@@ -56,7 +58,53 @@
 }
 
 
-class CvpApi(object):
+def sanitize_warnings(data):
+    ''' Sanitize the warnings returned after validation.
```
rewati-arista, line 62: a few typos in the docstring, correct version:

Sanitize the warnings returned after validation.

In some cases where the configlet has both errors
and warnings, CVP may split warnings that have
`,` across multiple strings.
This method concatenates the strings back into one string
per warning, and corrects the warningCount.

Args:
data (dict): A dict that contains the result
of the validation operation
Returns:
response (dict): A dict that contains the result of the
validation operation

cvprac/cvp_client.py
```diff
@@ -734,10 +740,8 @@ def _make_request(self, req_type, url, timeout, data=None,
                                ' response data. Attempt to decode')
                 decoded_data = json_decoder(response.text)
                 return {'data': decoded_data}
-            else:
-                self.log.error('Unknown format for JSONDecodeError - %s',
-                               err_str)
-                raise error
+            self.log.error('Unknown format for JSONDecodeError - %s', err_str)
```
rewati-arista, line 743: convert to f-string?
```suggestion
            self.log.error(f"Unknown format for JSONDecodeError - {err_str}")
```

rewati-arista (approved):
LGTM, minor f-string suggestions and questions about the logic for my clarity. Nothing major stopping a merge so approving it :-) feel free to implement/ignore the suggestions. 

Also did you use any tool/linter to convert all those strings to f-string format? that seems like some work to be done manually 😅

Makefile
```diff
@@ -49,7 +49,7 @@ coverage_report:
 
 pep8:
 	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE) cvprac/
-	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE),E402 test/
+	-pep8 -r --max-line-length=120 --ignore=$(PEP8_IGNORE),E402,E501 test/
```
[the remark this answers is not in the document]
mharista: In this case it only applies for the test directory but it is still a good point. I'll look into ways to resolve this as there is technically only one test file I want to ignore line length for.

cvprac/cvp_api.py
```diff
@@ -1,3 +1,5 @@
+# pylint: disable=fixme
+# pylint: disable=too-many-locals
```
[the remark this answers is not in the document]
mharista: Will move to one line as long as it doesn't hit line length cap.

cvprac/cvp_api.py
```diff
@@ -110,8 +158,7 @@ def cvp_version_compare(self, opr, version, msg):
                     running CVP version to.
         '''
         if opr not in OPERATOR_DICT:
-            self.log.error('%s is an invalid operation for version comparison'
-                           % opr)
+            self.log.error("{opr} is an invalid operation for version comparison")
```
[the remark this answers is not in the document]
mharista: I think you are technically right that we don't need f string formatting inside log message calls. I'll check to see if these actually triggered lint errors.

test/system/test_cvp_client_api.py
```diff
@@ -2127,8 +2131,7 @@ def test_api_inventory(self):
                                                       test_dev, orig_configlets,
                                                       create_task=True)
         # execute returned task and wait for it to complete
-        task_res = self.api.execute_task(results['data']['taskIds'][0])
-        self.assertEqual(task_res, None)
+        self.api.execute_task(results['data']['taskIds'][0])
```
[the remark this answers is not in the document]
mharista: This is removed because execute_task doesn't actually have a return statement and hence the linter was complaining. I can look into adding a return statement for execute_take as an alternative.

cvprac/cvp_client.py
mharista: I think from linting perspective we do not need to convert to f string within logging messages. I'll verify.

cvprac/cvp_api.py
mharista: Added back else and left return out of else.

mharista:
> LGTM, minor f-string suggestions and questions about the logic for my clarity. Nothing major stopping a merge so approving it :-) feel free to implement/ignore the suggestions.
> 
> Also did you use any tool/linter to convert all those strings to f-string format? that seems like some work to be done manually 😅

All comments addressed.

PR #275, opened 2024-04-29
Feat: Add support for searchTopology V3 endpoint

### LINKED ISSUE #257: modelName missing from searchTopology return data in 2023.1.0
Original searchTopology API has the modelName missing (field is just an empty string) as of CVP 2023.1.0.

There is a new API /v3/searchTopology that has the the modelName populated as expected. There are assorted other field differences between the two APIs return data as well.

mharista:
Fixes #257

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -2017,7 +2017,7 @@ class CvpApi():
             req_url = (f"/provisioning/searchTopology.do?queryParam={qplus(query)}&"
                        f"startIndex={start}&endIndex={end}")
         else:
-            # Newer CVP versions should use the V3 version of serach topology endpoint
+            # Newer CVP versions should use the V3 version of search topology endpoint
             req_url = (f"/provisioning/v3/searchTopology.do?queryParam={qplus(query)}&"
                        f"startIndex={start}&endIndex={end}")
 
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -1995,7 +1995,7 @@ class CvpApi():
             return self._save_topology_v2([])
         return None
 
-    def search_topology(self, query, start=0, end=0): # pylint: disable=too-many-branches
+    def search_topology(self, query, start=0, end=0):
         ''' Search the topology for items matching the query parameter.
 
             Args:
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -1995,7 +1995,7 @@ class CvpApi():
             return self._save_topology_v2([])
         return None
 
-    def search_topology(self, query, start=0, end=0):
+    def search_topology(self, query, start=0, end=0): # pylint: disable=too-many-branches
         ''' Search the topology for items matching the query parameter.
 
             Args:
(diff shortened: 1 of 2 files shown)


cvprac/cvp_api.py
```diff
@@ -2010,17 +2010,50 @@ def search_topology(self, query, start=0, end=0):
...
-                             f"startIndex={start}&endIndex={end}",
-                             timeout=self.request_timeout)
+        if self.clnt.apiversion is None:
+            self.get_cvp_info()
+        if self.clnt.apiversion <= 8.0:
+            # Original serach topology endpoint
+            req_url = (f"/provisioning/searchTopology.do?queryParam={qplus(query)}&"
+                       f"startIndex={start}&endIndex={end}")
+        else:
+            # Newer CVP versions should use the V3 version of serach topology endpoint
```
noredistribution, line 2020:
```suggestion
            # Newer CVP versions should use the V3 version of search topology endpoint
```

cvprac/cvp_api.py
```diff
@@ -2010,17 +2010,50 @@ def search_topology(self, query, start=0, end=0):
...
-                device['internalBuild'] = device['internalBuildId']
+
+                if 'isMLAGEnabled' in device:
+                    # original key was mlagEnabled but it changed to isMLAGEnabled
+                    device['mlagEnabled'] = device['isMLAGEnabled']
+                elif 'mlagEnabled' in device:
+                    # Key for V3 search topology changes back to mlagEnabled again
+                    device['isMLAGEnabled'] = device['mlagEnabled']
+
+                if 'isDANZEnabled' in device:
```
rewati-arista, line 2037: line 2037-2041 can be replaced by this, recently came across this which is the recommended pythonic way instead of adding conditional logic. 
```suggestion
               device['danzEnabled'] = device.get("isDANZEnabled", "")
```

cvprac/cvp_api.py
```diff
@@ -2010,17 +2010,50 @@ def search_topology(self, query, start=0, end=0):
...
+                    device['danzEnabled'] = device['isDANZEnabled']
+                else:
+                    # Key isDANZEnabled for V3 search topology is no longer in return data.
+                    device['danzEnabled'] = ""
+                    device['isDANZEnabled'] = ""
```
rewati-arista, line 2042: if we implement above change then this could change to:
```suggestion
                    device['isDANZEnabled'] = device['danzEnabled']
```

cvprac/cvp_api.py
```diff
@@ -2010,17 +2010,50 @@ def search_topology(self, query, start=0, end=0):
...
+
```
rewati-arista, line 2043: Similar suggestion for the next two keys: bootupTimestamp,  internalBuild as well 👍🏻

rewati-arista (approved):
lgtm, a small typo and suggestion (thanks to Gary who recently told me about https://docs.python.org/3/library/stdtypes.html#dict.get which prevents key errors without having to add a bunch of if else's)

cvprac/cvp_api.py
```diff
@@ -2010,17 +2010,50 @@ def search_topology(self, query, start=0, end=0):
                     netelement lists.
         '''
         self.log.debug(f"search_topology: query: {query} start: {start} end: {end}")
-        data = self.clnt.get(f"/provisioning/searchTopology.do?queryParam={qplus(query)}&"
-                             f"startIndex={start}&endIndex={end}",
-                             timeout=self.request_timeout)
+        if self.clnt.apiversion is None:
+            self.get_cvp_info()
+        if self.clnt.apiversion <= 8.0:
+            # Original serach topology endpoint
```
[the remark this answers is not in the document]
mharista: Fixed.

cvprac/cvp_api.py
mharista: Fixed.

cvprac/cvp_api.py
mharista: Fixed.

cvprac/cvp_api.py
mharista: Fixed.

PR #281, opened 2024-10-07
Fix: migrate to the new serviceaccount endpoints

noredistribution:


cvprac/cvp_api.py
```diff
@@ -3947,12 +3949,16 @@ def svc_account_token_get_one(self, token_id):
...
         msg = 'Service Account Resource APIs are supported from 2021.3.0+.'
-        if self.cvp_version_compare('>=', 7.0, msg):
-            payload = {"key": {"id": token_id}}
-            url = '/api/v3/services/arista.serviceaccount.v1.TokenService/GetOne'
-            self.log.debug(f"v7 {url} {payload}")
-            return self.clnt.post(url, data=payload)
-        return None
+        endpoint_legacy = '/api/v3/services/arista.serviceaccount.v1.TokenService/GetOne'
+        endpoint = '/api/resources/serviceaccount/v1/Token'
+        payload = {"key": {"id": token_id}}
```
mharista: Lets move the variable instantiations into the blocks they are used.

cvprac/cvp_api.py
```diff
@@ -3947,12 +3949,16 @@ def svc_account_token_get_one(self, token_id):
...
+        if self.cvp_version_compare('>=', 12.0, msg):
+            query_param = f"?key.id={token_id}"
+            self.log.debug(f'v12 {endpoint + query_param}')
+            return self.clnt.post(endpoint + query_param)
+        elif self.cvp_version_compare('>=', 7.0, msg) and self.cvp_version_compare('<', 12.0, msg):
```
mharista: We can remove the check for less than V12 since greater or equal is handled in the previous condition.

cvprac/cvp_api.py
```diff
@@ -4005,13 +4030,16 @@ def svc_account_get_all(self):
...
                       'time': '2022-02-10T04:28:14.251684869Z', 'type': 'INITIAL'}, ...]
-
         '''
         msg = 'Service Account Resource APIs are supported from 2021.3.0+.'
-        if self.cvp_version_compare('>=', 7.0, msg):
-            url = '/api/v3/services/arista.serviceaccount.v1.AccountService/GetAll'
-            self.log.debug(f"v7 {url}")
-            return self.clnt.post(url)
+        endpoint_legacy = '/api/v3/services/arista.serviceaccount.v1.AccountService/GetAll'
+        endpoint = '/api/resources/serviceaccount/v1/Account/all'
```
mharista: Move respective URL into its appropriate block.

cvprac/cvp_api.py
```diff
@@ -4026,11 +4054,16 @@ def svc_account_get_one(self, username):
                       'time': '2022-02-10T04:28:14.251684869Z'}]
         '''
         msg = 'Service Account Resource APIs are supported from 2021.3.0+.'
-        if self.cvp_version_compare('>=', 7.0, msg):
+        endpoint_legacy = '/api/v3/services/arista.serviceaccount.v1.AccountService/GetOne'
+        endpoint = '/api/resources/serviceaccount/v1/Account'
```
mharista: Move URLs into appropriate blocks.

cvprac/cvp_api.py
```diff
@@ -4070,8 +4103,13 @@ def svc_account_set(self, username, description, roles, status):
...
                                  'status': status}}
-            url = '/api/v3/services/arista.serviceaccount.v1.AccountConfigService/Set'
-            self.log.debug(f"v7 {url} {payload}")
+            if self.cvp_version_compare('<', 12.0, msg):
+                url = '/api/v3/services/arista.serviceaccount.v1.AccountConfigService/Set'
+                self.log.debug(f"v7 {url} {payload}")
+            else:
+                payload = payload["value"]
+                url = '/api/resources/serviceaccount/v1/AccountConfig'
+                self.log.debug(f"v12 {url} {payload}")
```
mharista: Is there a reason the conditionals are different here? Can we use the same format of check for >= 12 then fall through to checking >= 7?

cvprac/cvp_api.py
```diff
@@ -4087,9 +4125,14 @@ def svc_account_delete(self, username):
...
-            url = '/api/v3/services/arista.serviceaccount.v1.AccountConfigService/Delete'
-            self.log.debug(f"v7 {url} {payload}")
+            if self.cvp_version_compare('<', 12.0, msg):
+                payload = {"key": {"name": username}}
+                url = '/api/v3/services/arista.serviceaccount.v1.AccountConfigService/Delete'
+                self.log.debug(f"v7 {url} {payload}")
+            else:
+                payload = [{"name": username}]
+                url = '/api/resources/serviceaccount/v1/AccountConfig/deletesome'
+                self.log.debug(f"v12 {url} {payload}")
```
mharista: Same question as above with conditional format?

cvprac/cvp_api.py
mharista: This can be removed in all cases below as well.

cvprac/cvp_client.py
```diff
@@ -242,7 +242,13 @@ def set_version(self, version):
                               ' Appending 0. Updated Version String - %s',
                               ".".join(version_components))
             full_version = ".".join(version_components)
-            if parse(full_version) >= parse('2024.1.0'):
+            if parse(full_version) >= parse('2024.3.0'):
+                self.log.info('Setting API version to v14')
+                self.apiversion = 14.0
+            if parse(full_version) >= parse('2024.2.0'):
```
mharista: Update to `elif`

cvprac/cvp_api.py
```diff
@@ -3947,12 +3953,16 @@ def svc_account_token_get_one(self, token_id):
                       'time': '2022-05-03T15:38:53.725014447Z', 'type': 'INITIAL'}]
         '''
         msg = 'Service Account Resource APIs are supported from 2021.3.0+.'
+        payload = {"key": {"id": token_id}}
```
mharista: payload can be moved to the second conditional block.

mharista (approved):

PR #283, opened 2024-10-19
Fix: Move new svc_account resource APIs to v14 only

mharista:
Move new svc_account resource APIs to CVP 2024.3.0 v14 only due to bugs


noredistribution (commented):
Lgtm!

tgodaA (approved):
LGTM

PR #323, opened 2026-07-02
Feat: Allow cert-based login

vaibhavshaw-star:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement
- [ ] Version Bump

## Closes

Closes X

## Depends On

Depends on X

## Description

Added x509 cert-based login feature support. This allows users to pass client certificates while connecting to CVP nodes in case cert-based login feature is on.

## Testing

Added tests in the cvprac repo for the feature added and ran make tests.

## Impact

This adds support for client certificate login

## Effort required on reviewer's end
- [ ] Easy
- [x] Medium
- [ ] Hard

## Checklist
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/cvprac/cvp_api.py
+++ b/cvprac/cvp_api.py
@@ -194,6 +194,16 @@ class CvpApi():
             self.clnt.set_version(data['version'])
         return data
 
+    def get_cert_login_info(self):
+        '''Return the username associated with the certificate login cookie.
+
+            Raises:
+                CvpApiError: A CvpApiError is raised if the request is made
+                    without a valid certificate based authentication session.
+        '''
+        return self.clnt.get('/aaa/v1/certLoginInfo',
+                             timeout=self.request_timeout)
+
     # pylint: disable=too-many-arguments
     def add_user(self, username, password, role, status, first_name,
                  last_name, email, user_type):
--- a/cvprac/cvp_client.py
+++ b/cvprac/cvp_client.py
@@ -100,6 +100,7 @@ import logging
 from logging.handlers import SysLogHandler
 from itertools import cycle
 from packaging.version import parse
+from urllib.parse import urlparse, urlunparse
 
 import requests
 from requests.exceptions import ( # pylint: disable=redefined-builtin
@@ -114,6 +115,22 @@ from requests.exceptions import ( # pylint: disable=redefined-builtin
 from cvprac.cvp_api import CvpApi
 from cvprac.cvp_client_errors import CvpApiError, CvpLoginError, \
     CvpRequestError, CvpSessionLogOutError
+
+
+def url_with_port(url, port):
+    '''Return url with netloc port replaced by port.
+
+    For example, url_with_port('https://cvp.example.com:443/web', 9443)
+    returns 'https://cvp.example.com:9443/web'.
+    '''
+    parsed = urlparse(url)
+    host = parsed.hostname
+    if ':' in host and not host.startswith('['):
+        host = f'[{host}]'
+    netloc = host
+    if port:
+        netloc = f'{host}:{port}'
+    return urlunparse(parsed._replace(netloc=netloc))
 
 
 class CvpClient():
@@ -142,6 +159,9 @@ class CvpClient():
         self.apiversion = None
         self.authdata = None
         self.cert = False
+        self.cert_login = False
+        self.cert_login_port = None
+        self.client_cert = None
         self.connect_timeout = None
         self.cookies = None
         self.error_msg = ''
@@ -325,7 +345,8 @@ class CvpClient():
     def connect(self, nodes, username, password, connect_timeout=10,
                 request_timeout=30, protocol='https', port=None, cert=False,
                 is_cvaas=False, tenant=None, api_token=None, cvaas_token=None,
-                proxies=None):
+                proxies=None, cert_login=False, client_cert=None,
+                cert_login_port=9443):
         ''' Login to CVP and get a session ID and cookie.  Currently
             certificates are not verified if the https protocol is specified. A
             warning may be printed out from the requests module for this case.
@@ -366,6 +387,13 @@ class CvpClient():
                      Proxies can also be set via environment variables.
                      Please reference the below link for details of precedence.
                      https://requests.readthedocs.io/en/latest/user/advanced/#proxies
+                cert_login (boolean): Use CVP certificate based login flow.
+                    Supported on CVP 2026.3.0 and later. When True,
+                    client_cert must also be provided.
+                client_cert (str or tuple): Path to client certificate, or
+                    (cert, key) tuple as accepted by requests.
+                cert_login_port (int): CVP mTLS endpoint port. Default is
+                    9443.
 
             Raises:
                 CvpLoginError: A CvpLoginError is raised if a connection
@@ -388,6 +416,9 @@ class CvpClient():
                 nodes[idx] = os.environ.get('CURRENT_NODE_IP')
 
         self.cert = cert
+        self.cert_login = cert_login
+        self.cert_login_port = cert_login_port
+        self.client_cert = client_cert
         self.nodes = nodes
         self.node_cnt = len(nodes)
         self.node_pool = cycle(nodes)
@@ -538,6 +569,21 @@ class CvpClient():
             self.log.error(msg)
             raise CvpRequestError(msg)
 
+    def _validate_certificate(self):
+        '''Validate a client certificate for certificate based login.
+        '''
+        url = (url_with_port(self.url_prefix_short, self.cert_login_port) +
+               '/aaa/v1/validateCertificate')
+        response = self.session.get(url, headers=self.headers,
+                                    timeout=self.connect_timeout, verify=self.cert,
+                                    cert=self.client_cert)
+        self._is_good_response(response, f"Validate certificate: {url}")
+
+        if self.cookies is None:
+            self.cookies = response.cookies
+        else:
+            self.cookies.update(response.cookies)
+
     def _login(self):
         ''' Make a POST request to CVP login authentication.
             An error can be raised from the post method call or the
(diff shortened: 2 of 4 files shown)


cvprac/cvp_api.py
```diff
@@ -194,6 +207,44 @@ def get_cvp_info(self):
...
+                'cookies': response.cookies.get_dict()}
+
+    def get_cert_login_info(self):
+        '''Return the username associated with the certificate login cookie.'''
+        url = self.clnt.url_prefix_short + '/aaa/v1/certLoginInfo'
+        response = self.clnt.session.get(
+            url, cookies=self.clnt.cookies, headers=self.clnt.headers,
+            timeout=self.request_timeout, verify=self.clnt.cert)
+        self.clnt._is_good_response(response, f"Cert login info: {url}")
+        return response.json()
```
mharista: I think these should be moved to the `CvpClient` with other auth related code.

cvprac/cvp_api.py
```diff
@@ -102,6 +103,18 @@ def sanitize_warnings(data):
...
+def url_with_port(url, port):
+    '''Return url with netloc port replaced by port.'''
+    parsed = urlparse(url)
+    host = parsed.hostname
+    if ':' in host and not host.startswith('['):
+        host = f'[{host}]'
+    netloc = host
+    if port:
+        netloc = f'{host}:{port}'
+    return urlunparse(parsed._replace(netloc=netloc))
```
mharista: This should be moved to `cvp_client.py` since it is used in the auth related code.

cvprac/cvp_client.py
```diff
@@ -366,6 +370,12 @@ def connect(self, nodes, username, password, connect_timeout=10,
                      Proxies can also be set via environment variables.
                      Please reference the below link for details of precedence.
                      https://requests.readthedocs.io/en/latest/user/advanced/#proxies
+                cert_login (boolean): Use CVP certificate based login flow.
+                    When True, client_cert must also be provided.
```
mharista: I think we should add the CVP version where this feature support starts to the doc string.

cvprac/cvp_api.py
```diff
@@ -102,6 +103,18 @@ def sanitize_warnings(data):
     return data
 
 
+def url_with_port(url, port):
+    '''Return url with netloc port replaced by port.'''
```
mharista: An example of URL reformat would be useful in this docstring.

cvprac/cvp_client.py
```diff
@@ -602,14 +615,30 @@ def _login_on_prem(self):
...
         '''
         url = self.url_prefix + '/login/authenticate.do'
+        request_cert = None
+        if self.cert_login:
+            if self.client_cert is None:
+                raise CvpRequestError(
+                    'client_cert is required when cert_login is True')
+            self.api.validate_certificate(
+                self.client_cert, mtls_port=self.cert_login_port,
+                timeout=self.connect_timeout)
```
mharista: This should be refactored so this function is here in the client.

cvprac/cvp_api.py
vaibhavshaw-star: Done

cvprac/cvp_api.py
vaibhavshaw-star: Done

cvprac/cvp_api.py
vaibhavshaw-star: Done

cvprac/cvp_client.py
vaibhavshaw-star: Done

cvprac/cvp_client.py
vaibhavshaw-star: Done

cvprac/cvp_client.py
```diff
@@ -538,6 +569,41 @@ def _check_response_status(self, response, prefix):
...
+                                    cert=self.client_cert)
+        self._check_response_status(response, f"Validate certificate: {url}")
+
+        if self.cookies is None:
+            self.cookies = response.cookies
+        else:
+            self.cookies.update(response.cookies)
+
+        return {'location': response.headers.get('Location'),
+                'cookies': response.cookies.get_dict()}
```
mharista: Why are we returning this if we do not use it? We only check the result in one of the tests?

cvprac/cvp_client.py
```diff
@@ -538,6 +569,41 @@ def _check_response_status(self, response, prefix):
...
+    def _validate_certificate(self, timeout=None):
+        '''Validate a client certificate for certificate based login.
+
+            Args:
+                timeout (int): Optional request timeout.
+
+            Returns:
+                response (dict): Validation redirect location and cookies.
+        '''
+        timeout = timeout or self.connect_timeout
```
mharista: I think the timeout handling can be refactored. Since this is a private method and the only place internally it is called has a timeout hardcoded to the `self.connect_timeout` I think we can remove the timeout parameter entirely and simply use `self.connect_timeout` in the `.get` request.

cvprac/cvp_client.py
```diff
@@ -538,6 +569,41 @@ def _check_response_status(self, response, prefix):
...
+
+    def get_cert_login_info(self):
+        '''Return the username associated with the certificate login cookie.'''
+        url = self.url_prefix_short + '/aaa/v1/certLoginInfo'
+        response = self.session.get(url, cookies=self.cookies,
+                                    headers=self.headers,
+                                    timeout=self.api.request_timeout,
+                                    verify=self.cert)
+        self._is_good_response(response, f"Cert login info: {url}")
+        return response.json()
```
mharista: This method requires an already authenticated session so I think it can be moved to the API. Do you know how this will behave if you call it when not authenticated via cert?

cvprac/cvp_client.py
```diff
@@ -602,14 +668,28 @@ def _login_on_prem(self):
                     CVP node.  Destroy the class and re-instantiate.
         '''
         url = self.url_prefix + '/login/authenticate.do'
+        request_cert = None
```
mharista: I don't think we need to define this. We can use `self.client_cert` directly since it defaults to `None` as well.

cvprac/cvp_client.py
```diff
@@ -602,14 +668,28 @@ def _login_on_prem(self):
...
+        if self.cert_login:
+            if self.client_cert is None:
+                raise CvpRequestError(
+                    'client_cert is required when cert_login is True')
+            self._validate_certificate(timeout=self.connect_timeout)
```
mharista: Can drop timeout parameter as mentioned above.

mharista (changes requested):

cvprac/cvp_client.py
mharista: I see I mistakenly made the suggestion to move this to the client in my previous review comments. I now think it fits better in the API.

I am curious what the result is when calling this and not authenticated via cert?

cvprac/cvp_client.py
vaibhavshaw-star: Removed the return statement

cvprac/cvp_client.py
vaibhavshaw-star: Done

cvprac/cvp_client.py
vaibhavshaw-star: 1. Moved tp cvp_api.py
2. Throws error Session has expired.

cvprac/cvp_client.py
vaibhavshaw-star: Removed

cvprac/cvp_client.py
vaibhavshaw-star: Removed

cvprac/cvp_client.py
```diff
@@ -602,14 +648,26 @@ def _login_on_prem(self):
                     CVP node.  Destroy the class and re-instantiate.
         '''
         url = self.url_prefix + '/login/authenticate.do'
+        if self.cert_login:
+            if self.client_cert is None:
+                raise CvpRequestError(
+                    'client_cert is required when cert_login is True')
+            self._validate_certificate()
+            url = (url_with_port(self.url_prefix, 443) +
+                   '/login/authenticate.do')
```
mharista: This usage of `url_with_port` isn't needed and could even cause issues in some cases. The default `self.url_prefix` has `443` or user provided `self.port`. If a user provides a different port than 443 this will overwrite it to 443.

cvprac/cvp_client.py
```diff
@@ -538,6 +569,21 @@ def _check_response_status(self, response, prefix):
...
 
+    def _validate_certificate(self):
+        '''Validate a client certificate for certificate based login.
+        '''
+        url = (url_with_port(self.url_prefix_short, self.cert_login_port) +
+               '/aaa/v1/validateCertificate')
+        response = self.session.get(url, headers=self.headers,
+                                    timeout=self.connect_timeout, verify=self.cert,
+                                    cert=self.client_cert)
+        self._check_response_status(response, f"Validate certificate: {url}")
```
mharista: Is there a reason you are using `_check_response_status` here instead of `_is_good_response`?

cvprac/cvp_api.py
```diff
@@ -194,6 +194,16 @@ def get_cvp_info(self):
...
 
+    def get_cert_login_info(self):
+        '''Return the username associated with the certificate login cookie.'''
+        url = self.clnt.url_prefix_short + '/aaa/v1/certLoginInfo'
+        response = self.clnt.session.get(url, cookies=self.clnt.cookies,
+                                         headers=self.clnt.headers,
+                                         timeout=self.request_timeout,
+                                         verify=self.clnt.cert)
+        self.clnt._is_good_response(response, f"Cert login info: {url}")
+        return response.json()
```
mharista: This can be simplified to:

`return self.clnt.get("/aaa/v1/certLoginInfo", timeout=self.request_timeout)`

mharista (changes requested):

cvprac/cvp_client.py
mharista: Please add a note to the method doc string about the error being thrown if used without cert based auth.

vaibhavshaw-star: Added

cvprac/cvp_client.py
vaibhavshaw-star: Removed

cvprac/cvp_client.py
vaibhavshaw-star: we can use _is_good_response

cvprac/cvp_api.py
vaibhavshaw-star: yes, done.

mharista (approved):

